PR #224, opened 2017-07-03
Add Yubikey OTP Auth

di4bl0te:
Add Yubikey OTP Auth


src/install/hashtopussy.sql
```diff
@@ -560,7 +563,12 @@ CREATE TABLE `User` (
   `lastLoginDate`      INT(11)          NOT NULL,
   `registeredSince`    INT(11)          NOT NULL,
   `sessionLifetime`    INT(11)          NOT NULL DEFAULT '600',
-  `rightGroupId`       INT(11)          NOT NULL DEFAULT '1'
+  `rightGroupId`       INT(11)          NOT NULL DEFAULT '1',
+  `yubikey`            INT(1)           NOT NULL DEFAULT '0',
+  `otp1`               VARCHAR(50)      NOT NULL,
```
s3inlc: Is it correct that otp1 is varchar and the other three are ints?
As what I see from the other code, they should all have the same type.

src/index.php
```diff
@@ -29,6 +29,8 @@
         break;
       case '4':
         $message = "<div class='alert alert-warning'>You need to be logged in to view this! Please log in again.</div>";
+      case '':
```
s3inlc: Why the case here is empty?
As there is no break statement after the other case the OTP error message always will pop up when someone gets logged out. 
As the login procedure currently doesn't report why the login failed, maybe just include this error message into the failed login attempt (case 3).

src/inc/db.php
```diff
@@ -0,0 +1,10 @@
+<?php
```
s3inlc: db.php should not exist in the repository as it should be created from db.template.php either manually or by the installation process.

src/install/index.php
```diff
@@ -17,6 +17,7 @@
 use DBA\User;
 
 require_once(dirname(__FILE__) . "/../inc/load.php");
+require_once(dirname(__FILE__) . "/../inc/db.php");
```
s3inlc: The db files already get included by load.php, so this require is not necessary

src/install/hashtopussy.sql
djban: VARCHAR es el correcto para los 4 otp

src/inc/db.php
djban: ok, lo borro.

src/install/hashtopussy.sql
djban: Tengo que modificar varias cosas, no lo envies aun.

djban: Terminado @s3inlc.

s3inlc:
as mentioned in the referencing pull request, changes are included so this pull request is obsolete.

winxp5421:
Thank you @di4bl0te for your contribution.

PR #256, opened 2017-08-10
Update QueryFilterNoCase.class.php

niels1189:
MySQL isn't case sensitive unless specifically designed to be.
Not to mention that the LOWER() function is EXTREMELY slow especially with many entries.

This increases speed by a thousand fold

### DIFF
--- a/src/dba/QueryFilterNoCase.class.php
+++ b/src/dba/QueryFilterNoCase.class.php
@@ -32,7 +32,7 @@ class QueryFilterNoCase extends Filter {
       }
       return $table . $this->key . " IS NULL ";
     }
-    return "(LOWER(" . $table . $this->key . ") " . $this->operator . "? OR " . $table . $this->key . $this->operator . "?)";
+    return "($table . $this->key . $this->operator . "?)";
   }
   
   function getValue() {


s3inlc:
It's not that we want to just query something. Having removed the LOWER we would have exactly what we have in `QueryFilter.class.php`.
This class is there to cover a very specific case when we have searches which should be case sensitive and also when hashes are imported uppercase but Hashcat reports them in lowercase instead.

niels1189:
How is the choice between case and nocase designed?
Seeing as my hashtopussy was opting for NoCase and timing out constantly due to the slow queries.

s3inlc:
The problem is that people are uploading hashes in uppercase format, so when the client is reporting back with founds, they are in lowercase. So it has to check for both cases.

But we are planning to force all algorithms which are not case sensitive (MD5, SHA1, etc.) to be lowercases during import so we can remove this case insensitivity on the reporting on the api.

blazer2x:
If someone has time, they can go through https://hashcat.net/wiki/doku.php?id=example_hashes
and uppercase the hash and see whether hashcat cracks them and outputs them in lower form. If it does then filters can be setup to convert the hashes during import, note for case sensitive salts caution is required.

niels1189:
As I recall, only salts and  plains are considered case sensitive with hashes (hex) and if they are uploaded wrong, they can't be cracked.
Base64 encoded strings are of course case sensitive, but again. If uploaded wrong, they can't be cracked.

As I've never tested this, I can't exactly confirm this claim, but it'd make sense for hashcat to ouput the hash and salt in the exact format it was received in (upper and lowercase).
Even then, case insensitivity is something default in MySQL.

So perhaps it would be better to view what IS case sensitive and set the default to case insensitive to boost speed and performance.

s3inlc:
Maybe it could be an idea to let Hashcat doing the output in the format it received it. But I think this would be too complicated as I Hashcat stores such hashes as int numbers and therefore doesn't know anymore if the input was upper or lower hex.

And about the insensitivity of MySQL, this is very dangerous to assume that all have the default config values for this. For example the system where my test instance is running IS case sensitive and I cannot change this.

```
SELECT * FROM `Hash` WHERE hash like '%A3DB441655391D272361A6A43EE43Aa1%'
```
will not find the hash `A3DB441655391D272361A6A43EE43AA1`.

So the only solution to get rid of the case insensitive queries is to force to lowercase all import where possible.

NullPatrol:
I think this is related to charsets and collation.
i.e. utf8_general_ci is case insensitive.
https://dev.mysql.com/doc/refman/5.7/en/case-sensitivity.html

Do you happen to have the encoding and collation your database is using?

s3inlc:
Yeah you're right. My database is set to utf8_bin, so this is the reason for the case sensitivity.

But again, I think it can be critical when we assume that every user has a collation set which is case insensitive. And also if it happens that there are two different hashes which only differ in case this can cause trouble when the system is case insensitive.

NullPatrol:
That'd only be the case if two or more hashes differ in casing with a case senstive function AND they belong to the same hashlist. The chance this occurs is EXTREMELY small. Especially with low collision chance hashes. (which is almost all of them) Even then, they are rarely case sensitive.

MySQL323 for example is fairly prone to collisions.
But in essence is not case sensitive.
Perhaps even force the MySQL database to be case insensitive on install.

I recommend a default of non-case sensitive searching.
(Optional thought: Create an array of 'case sensitive' mysql corralations (or regex) and using a query determine if it's case sensitive or not)

My point - after all this rambling, is that case sensitive hashes are not uncommon.
But they are fetched by MySQL as they are put in. Identical hashes are extremely rare and as far as I'm concerned they are a real exception.

Putting the entire system at a huge disadvantage for a single exception is in my eyes a complete waste.
I'm not saying it's unneeded, rather the priority needs to be turned around.

s3inlc:
With the new upcoming version, this issue will be obsolete as the DB will be with the correct encoding and the case-insensitivity filters will not be used anymore for searches.

PR #259, opened 2017-08-10
New Telegram notification bot

NullPatrol:
Currently using the [[message]] as the others look bad.
Might require some info on how to configure the above message to suit needs (wiki)


s3inlc:
Wouldn't it be possible to use the receiver as the bot token which can be set when creating the notification?
This way it would not be required to edit anything in the source code and multiple bots can be used.

NullPatrol:
You require the 'chat' id between the bot and the user.
So one bot has multiple chats with the user.

s3inlc:
Ah yes, sorry I didn't saw this.
Looks good so far, I'll add more settings for new notifications later, so the bot token also can be set on the webpage.

Issue #301, opened 2017-11-23
Error adding new hashtype


BaptisteFortin:
Server version : __0.4.3__
Client version : __0.46.2__
Hashcat version : __4.0.1__

I'm trying to add a new hashtype for a custom hashcat kernel : 
- Hashtype : 16000
- Description : new_kernel
- Salted : No

I get the Error __` Failed to add new hash type! `__ but the hashtype is added into the list.
Same problem with any other hashtype name, description or with salted option "yes".

I tried to use it with an agent. The agent reports the password cracked but hashtopussy says it is not.

BaptisteFortin:
Adding a new hashtype works fine now, thank you.

I still have a problem when I try to use this new kernel I added.

My agent sends this report : 
{"action":"solve","token":"LND5oXkJMi","chunk":18,"keyspaceProgress":0,"progress":1000,"total":10000,"speed":8756,"state":5,"cracks":["FzRwXvJr2cS2AnZyNmO7lw==:YBhAKIzlvbPYi8fzqm6WKx0zwtbn4KDMe0E4Pvs5Eco=:1234"]}

The server responds like that : 
{"action":"solve","response":"SUCCESS",**"cracked":0**,"skipped":0,"zaps":[]}

s3inlc:
Is the hash saved to the database in the same format as it's printed when cracked? As there is a : separator there, is it a salted hash? 
Check in the `Hash` table if the hash was parsed correctly and the hash is in the hash column and the salt in the salt column.

BaptisteFortin:
I fixed my problem thank you.
The problem can from the fact that I did not check **`Salted hashes, separator`** in the hashlist creation form.

Why do you split hash and salt ? In hashcat everything is done in the parse function.

BaptisteFortin:
Problem solved thanks, the hashtype was not defined as salted.
Why do you split salt and hash ?

The kernels 1710, 5500, 5600, 8300, 8400, 8900, 10100, 10900, 11000, 11900, 12000, 12100, 12600, 13500, 13800, 15000 are not defined as salted, I think they should be.

All people using these kernels with hashtopussy might get **false negative result**.

winxp5421:
It's a rather large misconception that a salt should be part of the hash and I understand why most would think this. Both salt and hash are usually stored together in close proximity. However, the salt is part of the password and I think some people loose sight of this.

[...]

PR #305, opened 2017-12-12
fixed salted hashtype

BaptisteFortin:
#301


s3inlc:
Thanks for the fixes.

PR #340, opened 2018-03-06
Fix small tasks keyspace limit

imjustanoob:
Smalls tasks were given a limit of 1 instead of the entire keyspace.


s3inlc:
Can you please explain more detailed what this pull request should fix?

And why did you revert the latest commits on the master?

imjustanoob:
Small tasks were being broken into chunks with limit 1, because of the conditional error. 

The second commit, is me not knowing how to github.. not sure if you can ignore it or I have to do a repull.

s3inlc:
Ok thanks, I guess I see what the idea of the fix is.
If it's ok for you I'll just apply the change to master and then close this pull request, I think this is easier than try to revert the faulty merge here.

imjustanoob:
That works.

s3inlc:
Thanks for the contribution. Fix is pushed to master.

PR #411, opened 2018-08-01
Feature/prince

s3inlc:
Finished prince support.

There would be documentation needed in the wiki how to use prince tasks correctly.


s3inlc:
Added some first basic documentation for creating PRINCE tasks here: https://github.com/s3inlc/hashtopolis/wiki/Task-Creation-Guidelines

When we get feedback about the usage we can improve this later.

PR #421, opened 2018-08-06
Feature/agent info

s3inlc:
Agents can now send their gpu temperature and utility to the server where it can be monitored. Added status page where all agents can be viewed with their status.


s3inlc:
changing the base broke the checks, reopen in a new PR

PR #444, opened 2018-08-22
Testing framework

s3inlc:
Built test framework to run with Travis, one test is completed others to be added later.


s3inlc:
Still merging because the master at the moment is not ready to run because the default sql was not inserting one group correctly. Build should work after merge.

Issue #454, opened 2018-08-29
move pepper out of Encryption.class.php


cryptolovi:
Would be handy to have the peppers out of the above file and stored separate, so you don't have to manually update the file to include the right peppers on a new git fetch.

for example;
$PEPPER = file_get_contents(__DIR__ . “/.pepper”)
$PEPPER = file_get_contents(__DIR__ . “/.pepper1”)

file_put_contents() to write peppers

s3inlc:
Starting from the next release, all the specific values to set on installation are in `src/inc/config.php`, so no repository files will be changed on installation to avoid merging problems on updates.

cryptolovi:
Fantastic, much better solution also.

PR #461, opened 2018-09-03
Enhancement/install conf

s3inlc:
Moving all peppers out of actual files (Encryption and CSRF) to avoid merge conflicts on updates.


s3inlc:
Merge still can be done because it will be successful after merge, due to the wrong execution of the test with the base branch.

roycewilliams:
As suggested in #454 - nice idea, @cryptolovi

PR #473, opened 2018-09-27
Feature/health check

s3inlc:
Hashtopolis admins can trigger a health check which will require every registered agent to run a specific cracking test to test if they work correctly.


s3inlc:
build might've failed due to the wrong base branch at the beginning and also because of changed test,

PR #478, opened 2018-10-05
Enhancement/tests

s3inlc:
Including new testing of upgrade scripts starting from 0.8.0.


s3inlc:
Merge anyway because after the merge the build should succeed again. It is failing because of some drastic test changes between current-dev and the tests branch

PR #493, opened 2018-10-26
fix foreign key fails "configSectionId" bug in sql dump by adding a c…

DarkGigaByte:
…onfig section with id 6


s3inlc:
Could you please replace 'Network' with 'Multicast' as this would be the correct title for this section, as in the current-dev branch.

DarkGigaByte:
Done

s3inlc:
Thanks

PR #494, opened 2018-10-26
Enhancement/tests

s3inlc:
Containing completed preconfigured task test


s3inlc:
Still merging as it fails due to some changes. It will be pass tests again after merging.

PR #501, opened 2018-11-15
Feature/auto update

s3inlc:
Including routines to allow auto-updates at a later point.
Includes bundling functionality to easier distribute updates with the necessary files only.


s3inlc:
Build failed because a new file is not present on the base branch. Should succeed after merge.

PR #560, opened 2019-06-25
Current dev

s3inlc:
merge enhancements and bugfixes into master for release


s3inlc:
the pr build will fail due to the change in the test process where the order was not correctly on upgrade

Issue #565, opened 2019-06-26
Mask API Keys not belonging to a user


evilmog:
If you could enable an option to mask api keys not belonging to a user and have that set via a file on the fs and not via the webui it would be greatly appreciated.

Each api key has a user associated to it,  some environments have multiple admins and it would be handy to ensure that only the user to whom an api key belongs can see that api key in full.

s3inlc:
The feature is added in the current-dev branch, please close if this issue is resolved.

s3inlc:
I'm going to close the issue, the feature was tested and working. If this is not the case, please re-open.

PR #568, opened 2019-07-25
Mask API keys for other users with config setting

noncenz:
Addressing enhancement #565 this change masks viewing of other user's API keys by administrators. This change also prevents reassignment of API keys between users. 

Set `$MASK_API_KEYS = true` in conf.php to activate the feature.

### DIFF
--- a/src/account.php
+++ b/src/account.php
@@ -1,8 +1,11 @@
 <?php
 
 use DBA\Factory;
+use DBA\APiKey;
+use DBA\QueryFilter;
 
 require_once(dirname(__FILE__) . "/inc/load.php");
+
 
 if (!Login::getInstance()->isLoggedin()) {
   header("Location: index.php?err=4" . time() . "&fw=" . urlencode($_SERVER['PHP_SELF'] . "?" . $_SERVER['QUERY_STRING']));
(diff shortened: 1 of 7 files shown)


s3inlc:
Hey, sorry about the delay. Thanks for opening the PR providing this feature, some points to consider:
- The conf.php is not thought to put such config values in there, it's only for the installation setup (MySQL settings). If possible please use the same way as all other config variables are set (defined in src/inc/defines/config.php and created in the sql setup and the upgrade scripts)
- What is the way how admins give the api keys to their according users after creation if they cannot see it? (Note that normally users don't have full admin rights and therefore cannot see this section to get their own API key)
- The key owner change check should be executed directly in the action handler switch case and not outside of it.

noncenz:
Hi Sien, I think I'll need a little more guidance on this - I read the enhancement as the app instance having multiple admins and we don't want admins to see each other's API keys. 

- We can't expose the config setting to Admins through the GUI as they could just turn it off. Would it the best to add it to the database and just not map to the UI, or maybe use an include file other than conf.php?

- I had assumed all users were admins in this case and they would see their own keys through the API Management screen; an oversight on my part. Regular users will need to see keys without exposing them to an admin. Should we list them out under My Account > Settings? Or perhaps admins can see all user keys but other admin keys are masked?

s3inlc:
- You're right, didn't thought at that the admins could easily change the config. I will have to think about what the ideal solution would be to put that config. Maybe the conf.php would be good, but I would like to avoid global variables, maybe some other construct.
- Yeah, showing a users' own keys would be the best solution I think (at the location you proposed). This would also work for admins then, so all api keys on the config could be masked and admins also can see their own keys on their account page.

s3inlc:
Sorry about the big delay, I didn't had time so far to have a look at it. The code looks great, I'll merge it now and then maybe change some minor details later.
Thanks for the contribution!

PR #586, opened 2020-02-01
Tweeked Dark Mode UI

MilzoCSP:
The original dark css applies a rather large padding around buttons and boxes, font size is also a little smaller.
I've made changes to padding around buttons and boxes, increased font size by 1px with some minor colour changes on info panels, these changes now closely match the default UI.


s3inlc:
Thanks for the UI fixes

PR #598, opened 2020-02-10
Dark Mode UI Changes

MilzoCSP:
Darkened alerts, buttons and log entries highlighting.
Removed black borders from tables and replace with light a grey.
Added dark mode custom css file, where all future UI changes should go.


s3inlc:
Looks great, thanks again!

PR #617, opened 2020-03-04
update 0.6.0

H4xl0r:
??? or am i wrong


s3inlc:
You are right, must have overlooked this.

PR #621, opened 2020-03-26
add gitignore for working directories

nbirnel:
Running `git status` or `git ls-files` on a deployed instance of
hashtopolis is noisy, because working files show up as untracked.

This commit gitignores everything but .htaccess and .gitignore in

src/files/
src/import/
src/log/
src/tmp/


s3inlc:
Thanks 👍

PR #633, opened 2020-05-26
Allow monitoring of agent cpu utilisation data

SherlockNL:
Show average CPU utilisation data per agent on agent status page.
Display a graph of average CPU utilisation on the agent detail page.

Note that this requires an update of the hashtopolis-agent-python code in order for the agent to send the CPU utilisation data to hashtopolis. Import available psutil package in hashcat_cracker.py and add after line 315: 
`query['cpuUtil'] = psutil.cpu_percent()`
This change is proposed in a pull request: https://github.com/s3inlc/hashtopolis-agent-python/pull/10

### DIFF
--- a/src/agentStatus.php
+++ b/src/agentStatus.php
@@ -55,7 +55,7 @@ foreach ($stats as $stat) {
     $agentStats->addValue($stat->getAgentId(), $stat);
   }
 }
-UI::add('stats', $agentStats);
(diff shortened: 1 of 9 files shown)


s3inlc:
Thanks for the nice addition. I'll merge it now and change some smaller things to be consistent.

PR #636, opened 2020-06-16
hashcat 6.0.0 released

smarteam:
Since 6.0.0 is out of beta, maybe it's time to use it as main crackerbinary?


s3inlc:
indeed, thanks for the reminder.

PR #637, opened 2020-06-24
fix improper decrement calls in HashlistUtils::delete()

mlazeroff:
Currently, attempting to delete a hashlist which is part of a superhashlist does not complete properly and results in a blank page.

This fixes the error which was due to not providing a column key to Factory::getHashlistFactory->dec().


s3inlc:
Thanks, indeed this was wrong there.

PR #641, opened 2020-07-05
fix .map debug output for Chrome DevTools (quiet console)

roycewilliams:
This adds the appropriate JS "SourceMap" .map file for Popper, to quell warnings in Chrome DevTools and enable debugging of minified Popper.js
Also, switch to stock version of js-cookie 2.2.1 (https://github.com/js-cookie/js-cookie/releases/download/v2.2.1/js.cookie-2.2.1.min.js), which drops size from 1.9K to 1.6K and also drops the .map reference, quelling the other .map error in Chrome DevTools. (This was easier than trying to locate the .map file, which is referred to by the .js file with a relative path and a hash filename that is unlikely to be compatible with many installations (sourceMappingURL=/sm/b0ce608ffc029736e9ac80a8dd6a7db2da8e1d45d2dcfc92043deb2214aa30d8.map))


roycewilliams:
Note that the version of popper.min.js is dated 2018 and is close, but not an exact match, to version 1.14.3 as found in https://unpkg.com/browse/popper.js@1.14.3/. 

But by MD5 (83fb8c4d9199dce0224da0206423106f), it appears to be the version at https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js, so this .map file is from https://cdnjs.cloudflare.com/ajax/libs/popper.js/1.14.3/umd/popper.min.js.map.

s3inlc:
Thanks for the contribution, looks good.

PR #645, opened 2020-07-14
Fix spelling bug in menu template

SherlockNL:


s3inlc:
Oh my bad.. Just realized that this is not a mistake.. it's actually named like that, as lists_new is for normal hashlists and lists_snew for superhashlists.

PR #653, opened 2020-07-14
User-api / set task to top priority

SherlockNL:
Allow to set priority of tasks and supertasks to top priority within the set of pretasks. When setting the a task to be top priority, add one hundred to the previous highest priority and use this value.

Set task to top priority - protocol
```
 The clients sets a task to top priority.

 {
     "section": "task",
     "request": "setTaskTopPriority",
     "taskId": "1",
     "accessKey": <myKey>
 }

 The server should reply following on success:

 {
     "section": "task",
     "request": "setTaskTopPriority",
     "response": "OK"
 }
```
Set supertask to top priority - protocol
```
The clients sets a supertask to top priority.

{
    "section": "task",
    "request": "setSupertaskTopPriority",
    "supertaskId": "1",
    "accessKey": <myKey>
}

The server should reply following on success:

{
    "section": "task",
    "request": "setSupertaskTopPriority",
    "response": "OK"
}
```

### DIFF
--- a/src/inc/defines/tasks.php
+++ b/src/inc/defines/tasks.php
@@ -94,6 +94,9 @@ class DTaskAction {
   const SET_PRIORITY      = "setPriority";
   const SET_PRIORITY_PERM = DAccessControl::MANAGE_TASK_ACCESS;
   
+  const SET_TOP_PRIORITY = "setTopPriority";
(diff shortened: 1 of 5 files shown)


s3inlc:
Thanks for the contribution, I will update the User API documentation afterwards.

PR #664, opened 2020-09-23
Do not show hashlists that the user has no access to.

### LINKED ISSUE #661: Any user can view hashlists belonging to any access group

As long as the user has "can view hashlists" permission, it is possible to view any hashlist in the database by entering the hashlist id in the browser address bar (https://server/hashlists.php?id=X). The user does not have to belong to the same access group as the hashlist. When listing hashlists, only the expected hashlists are shown.

If you do the same thing for agents (listing or viewing one specific agent), it works as expected. It is not possible to view agents that do not belong to the same access group as you.

OneHumpCamel:
Fixes #661


s3inlc:
Thanks for the contribution!

PR #665, opened 2020-09-23
Users should not be able to view tasks with hashlists in other access groups

### LINKED ISSUE #662: Any user can view tasks belonging to any access group

As long as the user has "can view tasks" permission, it is possible to view any task in the database by entering the task id in the browser address bar (https://server/tasks.php?id=X). The user does not have to belong to the same access group as the task. When listing tasks, only the expected tasks are shown.

If you do the same thing for agents (listing or viewing one specific agent), it works as expected. It is not possible to view agents that do not belong to the same access group as you.

OneHumpCamel:
Fixes #662 

The code checked if all files in the task belongs to access groups that the user belongs to, but did not check if the user was allowed to view the hashlist. Added that check now.

### DIFF
--- a/src/tasks.php
+++ b/src/tasks.php
@@ -74,6 +74,10 @@ if (isset($_GET['id'])) {
   }
   
   $hashlist = Factory::getHashlistFactory()->get($taskWrapper->getHashlistId());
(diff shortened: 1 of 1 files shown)


s3inlc:
Thanks for the contribution!

PR #666, opened 2020-09-24
Issues/663 chunks should be filtered on access groups

### LINKED ISSUE #663: Entries in chunk activity should be filtered according to access groups

A user should not be able to see chunks assigned to agents not belonging to any of their access groups or chunks associated with a task not belonging to any of their access groups.

OneHumpCamel:
Potentially Fixes #663 

The solution for filtering out chunks where the task uses files that the user does not have access to violates the rule of not writing any SQL by hand, but I could not figure out any other way to solve it in a DB query. The filtering could of course be done in PHP code, but on a system with a large number of chunks, filtering in PHP may become rather slow.

### DIFF
--- a/src/chunks.php
+++ b/src/chunks.php
@@ -1,9 +1,12 @@
 <?php
 
+use DBA\AccessGroupUser;
 use DBA\Chunk;
+use DBA\ContainFilter;
 use DBA\OrderFilter;
 use DBA\JoinFilter;
 use DBA\Task;
+use DBA\TaskWrapper;
 use DBA\QueryFilter;
 use DBA\Factory;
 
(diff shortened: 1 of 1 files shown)


s3inlc:
Thanks for the contribution. Regarding the filtering there is one question, do we really need to hide the chunks where a user would not have access to the files, or wouldn't it make more sense, to show the filename, but not letting a user accessing that file if he does not have access to?
The same is done for example with agents which work on tasks, but the user does not have access to view or manage the agents. A user would just see the agent name, but it would not consist of a link.
Would you agree with that view?

OneHumpCamel:
> Thanks for the contribution. Regarding the filtering there is one question, do we really need to hide the chunks where a user would not have access to the files, or wouldn't it make more sense, to show the filename, but not letting a user accessing that file if he does not have access to?
> The same is done for example with agents which work on tasks, but the user does not have access to view or manage the agents. A user would just see the agent name, but it would not consist of a link.
> Would you agree with that view?

Agreed. As long as the user cannot view the file itself it should be fine. File names are not visible in the chunk list though, and I don't think you can view a task anymore unless you have access to all files used by it.

The code to filter out chunks where the user does not have access to the hashlist doesn't work correctly either, so we should put the entire PR on hold for now.

PR #672, opened 2020-10-23
New Task: Added Hashcat Command Validation

shivanraptor:
When creating a new task, the `hashcat` command is parsed and validated by [Optparse.js](https://github.com/shivanraptor/optparse-js) and Optparse Hashtopolis plugin. 

The task name is marked as required as well, and the total length of the command is checked (now limited to 200) to avoid database field truncation.


shivanraptor:
Added Optparse Hashtopolis Plugin version to 0.2 to add support of Combinator attack command validation.

shivanraptor:
Modified the display of "Estimated time" to display a number of years if days are over 365 days.

s3inlc:
Thanks for the contribution!
There are a few points on these changes. 
- First, it should distinguish if a user selected hashcat or another cracker program and only be active if hashcat was selected. 
- Ideally, as there are so many things possible with hashcat and it's maybe not always possible to keep up-to-date with all hashcat options, it would be good to have this feature as option in the server config (If you are not sure about this, I can do that after the merge, just let me know how I best should deactivate the functionality).
- Hashcat now has a new attack mode (-a 9), any plans of having it in the validation as well?
- And last but not least, remote files are a no-go. There are users not having their machines connected to the internet and also dynamically loading code from a third-party source would be suspect to quite some users. These files should be included locally.

Please let me know if there is anything not clear or if you are not sure how you should change some parts, I will try to help with that, or implement it afterwards.

shivanraptor:
Agree with your points and concerns. 

1. Yes, the changes will detect whether hashcat cracker is being used or not. If hashcat is not used, the validation should not be implemented.
2. Yes, it could be implemented as an option as server config. 
3. I am looking into hashcat v6.1.1, and I only see the following attack modes: 0, 1, 3, 6, 7
4. The validation JS files can be bundled or marked as an optional dependency of your next release.

I will try to work on the first 2 points. Hope I can help you to improve Hashtopolis.

s3inlc:
Thanks for looking into it.

Regarding the new attack mode. As far as I remember correctly, in 6.1.1 release this was not implemented, it was added since and is available on the master branch on the repository.

SherlockNL:
Correct me if I am wrong, but it seems the parser does not correctly parse attack commands that have multiple consecutive spaces in it. Hashcat does accept such commands.
This is simply fixed by applying `replace(/\s+/g, ' ')`

src/inc/Util.class.php
```diff
@@ -782,6 +782,11 @@ public static function sectotime($seconds) {
       $days = floor($seconds / 86400);
       $return = $days . "d ";
       $seconds = $seconds % 86400;
+      if($days > 365.25) { // taken leap year into consideration
+      	$years = floor($days / 365.25);
+      	$days = $days  % 365.25;
+      	$return = number_format($years) . "y " . $days . "d ";
```
TheWorkingDeveloper, line 788: Would this not yield the same results?
```
if ($seconds >= 86400*365.25) {
    return .= gmdate("Y", $seconds)."y ";
}
if ($seconds >= 31557600) {
    return .= gmdate("d", $seconds)."d ";
}
return $return.gmdate("H:i:s", $seconds);
```

src/templates/tasks/new.template.html
```diff
@@ -382,4 +383,37 @@ <h2>New task</h2>
     </div>
   </div>
 </form>
+<script src="https://cdn.jsdelivr.net/gh/shivanraptor/optparse-js@1.1.1/lib/optparse.js"></script>
+<script src="https://cdn.jsdelivr.net/gh/shivanraptor/optparse-js@1.1.1/lib/optparse.hashtopolis.js"></script>
```
TheWorkingDeveloper, line 387: external library might need to be made internal?

TheWorkingDeveloper (changes requested):

src/inc/Util.class.php
shivanraptor: There are typos. In both lines within the `if` clause, the `return` should be `$return` instead. No offence, but your suggested code works exactly the same as my code, isn't it?

src/templates/tasks/new.template.html
shivanraptor: Right, the author said so too. I'm going to update the pull request.

shivanraptor: Added the changes at commit 5e4b955: https://github.com/s3inlc/hashtopolis/pull/672/commits/5e4b955f4466bd9889ac7440cc16ff18763711d5

src/inc/Util.class.php
TheWorkingDeveloper: A lot less math is involved and it's in line with other code, using the same functions

TheWorkingDeveloper (approved):

PR #675, opened 2020-11-16
Issues/674 permission for managing access groups

### LINKED ISSUE #674: Separate permissions for managing access groups from managing users
- Your current Server version located at the bottom of any Hashtopolis webpage.
0.12.0

New feature: it should be possible for a user to manage access groups without also having permissions to manage users.

Today if you want to provide someone access to managing access groups it can only be done by giving the user permissions to manage users. That may not be what you want. If someone other than administrators have the responsibility to reassign people or agents between access groups, there needs to be a separate permission for managing access groups.

OneHumpCamel:
Fixes #674 

The _Users_ menu will be visible if the user has _Can manage users_ and/or _Can manage access groups_ permissions.

### DIFF
--- a/doc/user-api/sections/access.tex
+++ b/doc/user-api/sections/access.tex
@@ -79,7 +79,8 @@
     "addFileAccess": false,
     "crackerBinaryAccess": false,
     "serverConfigAccess": false,
-    "userConfigAccess": false
+    "userConfigAccess": false,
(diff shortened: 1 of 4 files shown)


s3inlc:
Thanks for the contribution!
I agree that it makes sense to have this separate for managing.

PR #680, opened 2020-11-24
Fix hashlist search percentage

SherlockNL:
add difference between chunk checkpoint and chunk skip to determine searched keyspace size


s3inlc:
Thanks!

PR #681, opened 2020-11-24
Enhancement/agentstatus data values

SherlockNL:
Show explicit data values for temperature and device/cpu utilisation per agent on agentstatus overview page
Previously only the color coding was shown, which holds less information.

Small bugfix: for assignments table on agentstatus page, remove redundant filter on chunk speeds to be added because statustimer is now variable


s3inlc:
Thanks for the contribution!

PR #683, opened 2020-12-20
Fix show cracks page

SherlockNL:
- Implement configurable number of hashes to be shown - this solves a bug where certain cracks are simply not listed

- Fix ordering on show cracks page

- Add pagination


cyrus104:
Thoughts on when this will be merged? Also if a merge is done what will need to be change in a running system to take advantage of this patch?

SherlockNL:
Merges are currently done by s3inlc. Nothing stops you from making use of this fix however!

Php and html files can be simply swapped while the system is live. The changes will take effect immediately. This code only changes what data is shown on the frontend. It is merely cosmetic, yet practical. 
To use this patch you could manually swap out the changed code files in the Hashtopolis source code with the updated version, or make use of Git.

s3inlc:
Thanks for the contribution, looks good!

PR #690, opened 2021-01-03
Hide info on data when there are no agents to show data on

TheWorkingDeveloper:
Before:
image
After:
image

### DIFF
--- a/src/templates/agents/status.template.html
+++ b/src/templates/agents/status.template.html
@@ -3,106 +3,109 @@
 <h2>Agents status ([[sizeof([[agents]])]])</h2>
 {%TEMPLATE->struct/messages%}
 <hr>
-<h3>Average device utilisation</h3>
-<div class="container-fluid">
-  <div class="row">
-    {{FOREACH agent;[[agents]]}}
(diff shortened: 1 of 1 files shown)


s3inlc:
I'm not sure if this is very useful. As it could be misleading to hide everything when there is no agent (people would not see what the page is for).

TheWorkingDeveloper:
The page is still used for showing agent status' a the graphs and an extra table is removed from view however when no agents are available and only the bottom 'agent activity' table is available for preview.
Alternatively showing the graphs but making sure they're displayed in full but empty would work as well. But the way they look now is not good.

s3inlc:
I'm not sure if I understand exactly what you mean.
The case that the tables are empty is only unless there is the first agent added, so I'm not sure what the benefit would be to hide it before. Also there is no agent activity until at least one agent is added, so basically for all the parts on the status page it is the same time at which there is something shown.

TheWorkingDeveloper:
It's a visual improvement for the layout of the site by removing irrelevant objects but keeping the bare minimum when no agents are present.

s3inlc:
If there are not any agents at all on the server, the agent status page is pretty much irrelevant anyway, I don't really see why then it should be kept to a bare minimum to improve anything visually.

zyronix (changes requested):
I don't really see any value in this, doesnt make sense to hide info if there are no agents as there should be agents to make hashtopolis usable at all. If there is a real value to this please elaborate more.

PR #693, opened 2021-01-15
check for empty hashlist for new task

vbalyasnyy:
Signed-off-by: Vasilyy Balyasnyy <v.balyasnyy@gmail.com>


s3inlc:
Thanks for the input, ideally this check should be done after trying to retrieve the hashlist as currently it also is not checked if the hashlist ID is valid. I adapted this in a commit after the merge, see here: b244dcdc21d56191be7148561a4eed50eb17ec5e

PR #702, opened 2021-04-06
Print normal error message on db connection error

### LINKED ISSUE #640: feature: explicit check for missing database
_- Your current Server version located at the bottom of any Hashtopolis webpage.
Latest GitHub_

If the database is entirely missing, a raw error is printed to the browser:
```
Fatal Error ! Database connection failed: SQLSTATE[HY000] [1049] Unknown database 'hashtopolis'
```
Low priority, but instead, it would be useful if a simple wrapper around the error could keep the error in the server theme, etc.

s3inlc:
closes #640

### DIFF
--- a/src/dba/AbstractModelFactory.class.php
+++ b/src/dba/AbstractModelFactory.class.php
@@ -4,6 +4,7 @@ namespace DBA;
 
 use MassUpdateSet;
 use PDO, PDOStatement, PDOException;
+use UI;
 
 /**
  * Abstraction of all ModelFactories.
@@ -872,7 +873,8 @@ abstract class AbstractModelFactory {
       if ($test) {
         return null;
       }
-      die("Fatal Error ! Database connection failed: " . $e->getMessage());
+      UI::printError(UI::ERROR, "Fatal Error! Database connection failed: " . $e->getMessage());
+      return null;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


src/dba/AbstractModelFactory.class.php
```diff
@@ -872,7 +873,8 @@ public function getDB($test = false) {
       if ($test) {
         return null;
       }
-      die("Fatal Error ! Database connection failed: " . $e->getMessage());
+      UI::printError(UI::ERROR, "Fatal Error! Database connection failed: " . $e->getMessage());
```
TheWorkingDeveloper, line 876: It might be worth to consider if $e->getMessage() can be omitted from the message but written to a log instead.
By doing so the error message stays more generic but can still be debugged (good practice) albeit optional.

TheWorkingDeveloper (approved):

src/dba/AbstractModelFactory.class.php
s3inlc: So far there is no file-based log on the server side which would be suitable for such messages.
I don't really see a big benefit from moving the error message to some extra location (which users again first need to be aware of).

zyronix: I would agree that a log file would be better, but we do not really have a server log file; so this is a nice solution as for now.

PR #710, opened 2021-05-15
Hashcat 6.2.0 new hashes

exel80:
Hey,

Hopefully, I update the correct files.

Here are new hash ids from latest Hashcat version 6.2.0
Also, can you double-check that I made correct choices when tagging "issalt" and "isslowhash"?


s3inlc:
Thanks for the pull request, the salted flag looks good so far. Regarding the slow hash, this was not documented well so far, I added the command to extract this information from the hashcat repo in the documentation readme for the future: https://github.com/s3inlc/hashtopolis/blob/master/doc/README.md
I will merge this PR now and adjust the slow hashes flag for these.

PR #720, opened 2021-09-11
Adding counter for repeating devices on agents page and detail page

zyronix:
Small change to the agent overview page and agent detail page. If a device is repeating (for example you have 4 Nvidia 1080's in a agent, they will now be displayed as 4x Nvidia 1080 instead of 4 lines.

Before:
agent_count_2

After:
agent_count_1


s3inlc:
I think it may look even better without the single quotes around the multiplication part. What do you think?

PR #728, opened 2021-10-19
Allow Agents to be unassigned from their task.

TheWorkingDeveloper:
A bug causes $taskId to be empty / unset instead of 0 causing it to error saying no task exists.
This change fixes that bug.


zyronix:
Could you describe a bit more how you encounterd this? I am not able to create a situation where I ran into this line being a problem.

TheWorkingDeveloper:
> Could you describe a bit more how you encounterd this? I am not able to create a situation where I ran into this line being a problem.

Assign agent to a task and let it run. Attempt to unassign an agent from that task directly. Receive error: no such task exists or similar.

zyronix (approved):

PR #738, opened 2022-02-07
Updated LDAP Integration

### LINKED ISSUE #316: LDAP Authentication
Implement LDAP Authentication for single sign on in enterprise environments

gentoo9ball:
I've taken the older LDAP Pull Request and merged it with the latest code. It is working great in my environment.

From the old code, I've found that the popular kpeiruza/hashtopolis docker image uses the User.class constructor with 16 arguments, the extra argument breaks this. So I've reworked the constructor flow to use either 16 or 17 arguments.

Also, I've made the BaseDN and UID config settings, so this can work in different environments. Some LDAP Servers use CN by default, others use UID. Setting these variables right now has to occur manually via the database.

A couple nice things that I can work on in the future, would be:
1. A way to specify TLS Certs for LDAPS
2. A configuration interface to set LDAP Configuration settings (ldap_server, ldap_basedn, ldap_uid)
3. A way to sync users from LDAP given an admin user


gentoo9ball:
Still using this on our side, working well

s3inlc (changes requested):
Thanks for the pull request, I added some small comments regarding some code parts, but looking good overall. 
One thing which is missing, is the addition to the update script which is used to update existing installations which pulled new changes from the repository. These would break with the current implementation. The changes needed to be done on the database should be added in `src/install/updates/update_v0.12.x_v0.x.x.php`.

@zyronix will do a test on our side to verify the functionality and then report back.

Prototype82:
Is there any progress on this topic? LDAP authentication would be a great feature

zyronix:
No progress on this has been made, but in the background we have been working on a new frontend and backend. This will come with the switch to oAuth for authentication, which would make it possible to use all kinds of authentication servers like LDAP or AD based servers for authentication.

jasonCar87:
@gentoo9ball Could you tell me please what kind of account I need to create in ldap to enter its name as ldap_uid. I created an ldap service account with a password, but I didn't find any input field for that password in the hashtopolis app.

s3inlc:
As we now finally could switch to the new API after long term development, LDAP authentication should be possible via OAuth. If you test it out and encounter issues, please report them via the issues and we will be looking at it.

PR #742, opened 2022-02-24
Enhancement: increase size of value column for agentstats

zyronix:
In some cases the value size is not big enough to store the agents stats, resulting in an internal server error. This fixes that issue.

```
# php update.php

mysql> DESCRIBE AgentStat;
+-------------+--------------+------+-----+---------+----------------+
| Field       | Type         | Null | Key | Default | Extra          |
+-------------+--------------+------+-----+---------+----------------+
| agentStatId | int(11)      | NO   | PRI | NULL    | auto_increment |
| agentId     | int(11)      | NO   | MUL | NULL    |                |
| statType    | int(11)      | NO   |     | NULL    |                |
| time        | bigint(20)   | NO   |     | NULL    |                |
| value       | varchar(128) | YES  |     | NULL    |                |
+-------------+--------------+------+-----+---------+----------------+
5 rows in set (0.00 sec)
```

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -21,7 +21,7 @@
 - Adjusted handling to be able to deal with changed mode 22000 output.
 - Fixed pagination of hashes on cracks page.
 - Time of Zaps inserted is now saved.
-- Fixed unable to unassign agent from the task detail screen
+- Fixed unable to unassign agent from the task detail screen.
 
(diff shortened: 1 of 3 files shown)


soxrok2212:
Also a good one!

s3inlc (approved):

PR #743, opened 2022-02-24
Speed graph fixed with different status timer

zyronix:
To reproduce this bug, create a new task, set the status timer of the task to something different then the servers default. In my example: 30s (server default is 5s). Let the task run for a couple of status updates intervals. You'll see that the graph of the speed will contain drops and is not a consistent line. This is just because the graph is created with the servers default status timer in mind and creates data-points for status updates that are simply not there. This fixes that issue and makes sure the graph will be displayed correctly: a stable line in my example.

Before:
graph_before

After:
graph_after


soxrok2212:
This is a good one!

s3inlc (approved):

PR #744, opened 2022-02-24
Agent details hide/show button

zyronix:
This pull requests adds a hide/show button on the agent details page so that the graph and other information is easier to access.

Before:
agent_before

After:
agent_after_1

Clicking the button:
agent_after_2


zyronix:
@s3inlc requested to move the button inside the agent id cell.

zyronix:
Implemeting feedback, I moved the button out of the first table. Simply because it just looked ugly if it was in the same collum as the button.

Closed:
details_1

Opened:
details_2

s3inlc (approved):

PR #754, opened 2022-04-08
Visual Studio Code Devcontainer

zyronix:
This pull requests adds a visual studio devcontainer, which makes it easier to setup a development environment for hashtopolis-server. Including php debugging, breakpoints etc and detailed error logging. Aswell as easy switching between branches. See https://code.visualstudio.com/docs/remote/containers for more information

### Setup

1. Install vscode https://code.visualstudio.com/
2. Optional (windows): Install wsl2 https://docs.microsoft.com/en-us/windows/wsl/install
3. Install Docker Desktop https://www.docker.com/products/docker-desktop/
4. Install remote containers extension in vscode https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers
5. Optional (windows): Install remote wsl extension in vscode https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-wsl
6. Clone the server repo (and for now select this branch ofc)
7. In vscode select 'Open Folder in Container' either by (CTRL + SHIFT + P, type: Open Folder in Container) or press the green extension button in the left bottom and select the option, now select the folder where hashtopolis-server is
8. Optional: set a break point and press F5 (Run -> Start Debugging)
9. Open a browser and visit: http://localhost:8080/
10. Login in with: username: root, password: hashtopolis


s3inlc:
The pull request looks fine from my side. Maybe these setup instructions should be also placed into a wiki page so people later can still easily find it.
Also a open question would be the handling of the .devcontainer and .vscode folders on setups. Should we include the deletion of these in the install script which is used during normal setups? So these directories would not be exposed on a webserver?

zyronix:
Regarding the instruction: yes when it is merged I'll moved them to a wiki page. Regarding the .devcontainer folder, I have added a htaccess file, because the folder contained a php file so potentially this could have caused issues. This should prevent access to this directory. Regarding the .vscode, as far as I know we do not delete any other folder ourselfs right? For example the .github also is let intact? I would not see any reason to approach the vscode folder any different.

s3inlc (approved):

PR #755, opened 2022-04-09
Add new hashtypes

its5Q:


s3inlc:
Thanks for the updated list.

Regarding the 'isSalted' setting, I identified three modes, where it should be set to 1, these are 27800, 27900 and 28000. Not because the inputs there necessarily are salts, but because they are implemented in Hashcat (in their respective module C files) to use the `hashconfig->separator` config and not a hard-coded colon character. Therefore, it must be marked as 'isSalted' because Hashcat then prints the hash using the character given to the -p parameter (which is different than the colon in Hashtopolis).

Also, as of today, the mode 28900 is the newest one in the master branch of the repository of Hashcat (it seems their wiki page with the example hashes is not completely up-to-date with these). The question is, if these should also just be included as they sooner or later will be in the next release as well. And quite some people are using the Hashcat beta builds by default.

Last but not least, you updated the hash type entries for the initial installation setup. To also provide existing installations with the new hash types, the update script ideally also would be adjusted, i.e. in this file: `src/install/updates/update_v0.12.x_v0.x.x.php`.

its5Q:
Thanks for your explanation, I've added all modes from master and marked respectable modes as salted. I will now adjust the update script

its5Q:
I think everything is done

s3inlc:
Thanks for your adjustments, that looks fine.

PR #758, opened 2022-04-12
Notifications do not honor access settings

### LINKED ISSUE #757: Notifications are sent to user without considering access
Current server version: 0.12.0-repository

Any user can request to be notified on events like "hashlistCrackedHash" or "agentError", and can request to recieve notifications for any object. The user will then receive notifications for all objects in the system, not only the ones the user has access to.

Before the systems sends any notification, it should verify that the recipient has access to the object that triggered the notification.

OneHumpCamel:
Fixes #757

Check receiver permissions before sending notifications.


zyronix:
Thanks for your contribution, I can confirm the bug and see that indeed you have fixed the issue. I am not a export on the notification section, @s3inlc could you maybe take a look at it? If not I have to dive more into this.

s3inlc:
Thanks a lot. I was able to test and confirm that it works correctly now.

s3inlc (approved):

PR #764, opened 2022-05-18
Add support for maximum amount of agents that should work on a task.

cc-nfi:
This is a proposal implementation for limiting the maximum amount of agents working on a task, allowing the surplus of agents to work on tasks with lower priority. It is currently implemented for tasks only, the idea is that if this gets approved (after any feedback and changes), we will implement it for super tasks as well.

We added the possibility to configure the maximum amount of agents at a couple of location in the UI. Because the task overview table would become too large, we made a change by removing the files column, but showing if they are linked to any secret ones by rendering the padlock behind the task name. See the pictures below.

create-task
task-index
supertask-info

### DIFF
--- a/ci/tests/TaskTest.class.php
+++ b/ci/tests/TaskTest.class.php
@@ -50,6 +50,7 @@ class TaskTest extends HashtopolisTest {
       "crackerVersionId" => 0,
       "files" => [],
       "priority" => 0,
+      "maxAgents" => 0,
       "preprocessorId" => 0,
       "preprocessorCommand" => "",
       "accessKey" => "mykey"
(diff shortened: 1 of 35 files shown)


ci/tests/TaskTest.class.php
```diff
@@ -50,6 +50,7 @@ public function testCreateTask($values = [], $assert = true) {
       "crackerVersionId" => 0,
       "files" => [],
       "priority" => 0,
+      "maxAgents" => 0,
```
cc-nfi, line 53: Should also add this to the create pretask API request in PretaskTest

TheWorkingDeveloper (approved):
source code looks good, functionality not tested.

s3inlc (approved):

s3inlc:
Thanks, looks good and tested that it works!

PR #778, opened 2022-06-04
Added column "Created by" to hashlist overview

dr0pd34d:
As a group of people using the tool we always wondered where to see which user created the hashlist in order to provide support or give advice.
I thought I would try my best in order to add the feature through a pull request.

For now it is only shown in the "Hashlists" overview as a sortable column.
As so far the ID of the creating user has not been saved with the hashlist, I created the column in the database setup file.

I have no experience with the database migration scripts so please help me out here, thank you. :)

I hope everything is in the right spot and the solution to pass the creator username into the template is done in the right way.
Please let me know as I am always open for improvements.

Thank you for the great tool!

### DIFF
--- a/src/dba/models/Hashlist.class.php
+++ b/src/dba/models/Hashlist.class.php
@@ -17,8 +17,9 @@ class Hashlist extends AbstractModel {
   private $notes;
   private $brainId;
   private $brainFeatures;
+  private $creatorId;
   
-  function __construct($hashlistId, $hashlistName, $format, $hashTypeId, $hashCount, $saltSeparator, $cracked, $isSecret, $hexSalt, $isSalted, $accessGroupId, $notes, $brainId, $brainFeatures) {
+  function __construct($hashlistId, $hashlistName, $format, $hashTypeId, $hashCount, $saltSeparator, $cracked, $isSecret, $hexSalt, $isSalted, $accessGroupId, $notes, $brainId, $brainFeatures, $creatorId) {
     $this->hashlistId = $hashlistId;
     $this->hashlistName = $hashlistName;
     $this->format = $format;
@@ -33,6 +34,7 @@ class Hashlist extends AbstractModel {
     $this->notes = $notes;
     $this->brainId = $brainId;
     $this->brainFeatures = $brainFeatures;
+    $this->creatorId = $creatorId;
   }
   
   function getKeyValueDict() {
@@ -51,7 +53,8 @@ class Hashlist extends AbstractModel {
     $dict['notes'] = $this->notes;
     $dict['brainId'] = $this->brainId;
     $dict['brainFeatures'] = $this->brainFeatures;
-    
+    $dict['creatorId'] = $this->creatorId;
+
     return $dict;
   }
   
@@ -121,6 +124,14 @@ class Hashlist extends AbstractModel {
   
   function getCracked() {
     return $this->cracked;
+  }
+
+  function getCreatorId() {
+    return $this->creator;
+  }
+
+  function setCreatorId($creatorId) {
+    $this->creatorId = $creatorId;
   }
   
   function setCracked($cracked) {
@@ -197,4 +208,5 @@ class Hashlist extends AbstractModel {
   const NOTES = "notes";
   const BRAIN_ID = "brainId";
   const BRAIN_FEATURES = "brainFeatures";
+  const CREATOR = "creatorId";
 }
--- a/src/dba/models/HashlistFactory.class.php
+++ b/src/dba/models/HashlistFactory.class.php
@@ -23,7 +23,7 @@ class HashlistFactory extends AbstractModelFactory {
    * @return Hashlist
    */
   function getNullObject() {
-    $o = new Hashlist(-1, null, null, null, null, null, null, null, null, null, null, null, null, null);
+    $o = new Hashlist(-1, null, null, null, null, null, null, null, null, null, null, null, null, null, null);
     return $o;
   }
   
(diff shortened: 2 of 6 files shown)


s3inlc:
Thanks a lot for the contribution, it already looks pretty great.

As you mentioned, the creator is shown on the hashlist overview. I think it would be beneficial to also have it listed on the hashlist details page (e.g. below the "Member of Access Group" row), could you add this?

Regarding your question for the update script. The idea is that an atomic change in the DB should be wrapped in an if statement which checks for a unique key (to determine if it already was applied or not), e.g. `v0.12.x_maxAgents_pretask_task`. Inside this statement, in your case you would need to check if the column already exists and if not, create it (The changes which are done on the update with the above key should pretty much show what you would need as well). 
The only challenge left then, is that you need to take care about already existing hashlists. If there is no creator set for the existing tasks, it will result in internal server errors. I would propose to just set all of the existing tasks to the oldest (i.e. lowest) user ID, as this typically would be the admin user. 
Ideally the entry in the changelog (`doc/changelog.md`) is included in the PR itself, so please just add a short sentence about this new feature there.

Would it be possible to add these changes? I will test it more in detail afterwards and happily merge it.

dr0pd34d:
Thank you for the valuable feedback.

Sure, I will try my best to integrate it into the detail template too and of course in the change log and try my best with the database migration script.

I will report back ASAP

dr0pd34d:
Okay changes done, please check, I had to merge upstream commits so I hope that was done correctly :)

src/install/updates/update_v0.12.x_v0.x.x.php
```diff
@@ -187,3 +187,9 @@
   }
   $EXECUTED["v0.12.x_maxAgents_pretask_task"] = true;
 }
+if (!isset($PRESENT["v0.12.x_creatorColumn"])) {
```
zyronix, line 190: This code is still missing setting the creator field for existing hashlist.

src/dba/models/Hashlist.class.php
```diff
@@ -122,6 +125,14 @@ function setSaltSeparator($saltSeparator) {
   function getCracked() {
     return $this->cracked;
   }
+
+  function getCreatorId() {
+    return $this->creator;
```
zyronix, line 130: Should be `$this->creatorId;`

src/dba/models/Hashlist.class.php
```diff
@@ -197,4 +208,5 @@ function setBrainFeatures($brainFeatures) {
   const NOTES = "notes";
   const BRAIN_ID = "brainId";
   const BRAIN_FEATURES = "brainFeatures";
+  const CREATOR = "creatorId";
```
zyronix, line 211: should be `const CREATOR_ID = "creatorId";`

src/templates/hashlists/detail.template.html
```diff
@@ -20,6 +20,10 @@ <h2>Hashlist details</h2>
         <td>Member of Access Group:</td>
         <td>[[accessGroup.getGroupName()]]</td>
       </tr>
+      <tr>
+        <td>Creator:</td>
```
zyronix, line 24: Suggestion: make this also: 'created by' to make it consistent with the hashlist overview

zyronix (changes requested):
Small change, add the new field to the generator.php found in the src/dba/models/

```
$CONF['Hashlist'] = [
  'hashlistId',
  'hashlistName',
  'format',
  'hashTypeId',
  'hashCount',
  'saltSeparator',
  'cracked',
  'isSecret',
  'hexSalt',
  'isSalted',
  'accessGroupId',
  'notes',
  'brainId',
  'brainFeatures',
  'creatorId'
];
```

Bigger issue to resolve: if a user gets deleted; the creatorId should be reset of all of the issues owned by the user or else it will return an error when trying to list access a hashlist because it ID does not resolve to a username anymore.

dr0pd34d:
Thank you for looking over it, you are completely right, I forgot to add and cover those cases. 
I will report back once I made some progress.

src/dba/models/Hashlist.class.php
dr0pd34d: Fixed it here:
https://github.com/hashtopolis/server/pull/778/commits/4f004e63d7ee477285a67aa027c215ce3d03c8b6

My bad, forgot to stash the change, was already done locally.

src/dba/models/Hashlist.class.php
dr0pd34d: Fixed in https://github.com/hashtopolis/server/pull/778/commits/31c8787c8a5766d6ff0a9c989c4ae9b99037e554

src/templates/hashlists/detail.template.html
dr0pd34d: Fixed in https://github.com/hashtopolis/server/pull/778/commits/5892721e6e9f5b10efee6a3683f1f99194c8a106

zyronix:
@dr0pd34d How far are you on this? Is it ready to be reviewed again?

dr0pd34d:
As you can see I did resolve a lot of the requested changes.
For one change I added a question if it is implemented in a correct way.

The biggest issue was not tackled yet, which is:
`Bigger issue to resolve: if a user gets deleted; the creatorId should be reset of all of the issues owned by the user or else it will return an error when trying to list access a hashlist because it ID does not resolve to a username anymore.`

zyronix:
Do you need some help with implemnting that or are you working on it?

dr0pd34d:
@zyronix Yes please some help would be nice. Currently missing the time to follow up on the issue.

ObsidianOracle:
Will be implemented in the new ui. This PR adresses the old ui.

PR #786, opened 2022-06-11
Add queryFilter to tasks for only showing assignable agents for hashlists with specific AccessGroups

### LINKED ISSUE #777: Assign Agents to Task show all Agents, even though the list should be filtered by User-Access Group and Hashlist-Access Group
## Description
When creating an Access Group with for example 4 Agents and you give a user Access to this group, he sees only these 4 Agents in the `Agent Status` view. However, when creating a Task for a Hashlist, this user can select way more Agents than suspected.

## What is expected
It is expected that when the User access for agents and/or the Hashlist access for agents is restricted, the list of assignable agents is filtered according to the agents which can be assigned to the task with the specific hashlist.

## Reproducition steps
1. Create a new Access Group
2. Add at least 1 agent but not all of them into the new Access Group
3. Add your user to this Access group and make sure he is not in the `Default Group`
4. Create a new Hashlist with the freshly created Access Group
5. Create a new Task for that new Hashlist and click on the newly created Task
6. Under `Assigned agents`, click the Dropdown List. you should see something similar to the following image:

Agents-list-which-should-be-filtered

## MISC Data
- Your current Server version: Hashtopolis: 0.12.0+repository

robinduerhager:
Closes #777 .

In my environment i have 2 Access Groups. `Default` has no Agent, `restricted` has one Agent. There are also 2 Hashlists and 2 Tasks for each AccessGroup respectively. Here is the List of the assignable Agents for each Task:

In this Task it is expected that 1 Agent is shown since the linked hashlist uses the `restricted` Access Group with the one Agent.
restricted task one agent

In this Task it is expected that no Agent is shown since the linked hashlist uses the `Default` Access Group with no Agent inside.
default task no agent

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -28,6 +28,7 @@
 - Fixed sending two to headers when sending emails (issue #751).
 - Fixed access group not being changed on Hashlist detailed screen (issue #765).
 - Fixed missing check on permissions for sending notifications (issue #757).
+- Fixed unassignable agents are shown as assignable (issue #777).
 
 ## Enhancements
 
--- a/src/tasks.php
+++ b/src/tasks.php
@@ -1,5 +1,6 @@
 <?php
 
+use DBA\AccessGroupAgent;
 use DBA\Agent;
 use DBA\Assignment;
 use DBA\Chunk;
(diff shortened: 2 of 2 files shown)


zyronix (changes requested):
Looks good, could you please add a line to the change log referring your issue number?

robinduerhager:
Done :)

zyronix (approved):

PR #789, opened 2022-06-29
Failed to delete hashlist containing cracked hashes while agent is running on the hashlist

zyronix:
I was able to reproduce it:

-     Create a hashlist with multiple hashes
-     Create a task
-     Upload one crack for one of the hashes
-     While task is running, delete the task
-     Delete the hashlist boom


zyronix:
Added changelog entry. Should be good to go.

s3inlc (approved):

PR #792, opened 2022-07-26
Hashlists can be archived

s3inlc:
Allow archiving of hashlists. This does not stop any tasks which may still be running at this points, and also it does not archive any task belonging to the hashlist. But there can no new tasks be created using archived hashlists.
This allows a better overview over active hashlists without having to delete old hashlists.

Note: squash on merge


zyronix (commented):
If a task is still active, you can already archive a hashlist. I would say that either archving a hashlist would stop all tasks or tells you that there are still active tasks.

zyronix:
When you archive a hashlist you can still create a task through the API.

zyronix (changes requested):
Code works, but there is no check in the taskhandler to verify that when you create a task that the hashlist is active. Imo that should be doen, but I am open to discuss this.

zyronix (approved):

Issue #793, opened 2022-07-26
maxAgents value is not documented in User API, but required


s3inlc:
When sending a createTask request to the User API, it fails when no `maxAgents` value is provided, but this is not documented.
Also `maxAgents` could be an optional value and not required and would default to 0 (no limit).

PR #798, opened 2022-07-26
check for non-required values in user API createTask command

s3inlc:
closes  issue #793


zyronix (changes requested):
During discussion with @s3inlc the issue was raised in the API. The error was something like: cannot set value, because the database does not have a default value for maxAgent. 
My suggestion would be to make it consistent with Task_priority, which is also an optional value. Thus changing the maxAgent to be default 0 in the database.

PR #803, opened 2022-08-12
Docker build include git commit if possible

s3inlc:
When building a docker image for Hashtopolis, there will be an issue when the image later is updated and the container restarted. In such a case Hashtopolis is not able to figure out, that it was updated, as the build information remains set to `repository`.

In order to allow building docker containers which can later be updated and Hashtopolis being able to apply upgrades, it must include the git commit hash somewhere. This can be achieved by running `./build.sh` which will place the hash into a file which is then included in the docker build process.


s3inlc:
This PR is becoming obsolete with #866.

PR #804, opened 2022-08-19
Update hashtopolis.sql to fix yubikey_url.

bkuzmacissp:
The default yubikey_url for new installations was being set to a HTTP endpoint that has been deprecated by Yubikey. There was a previous migration from v0.3.2->v0.4.0 that updated this to HTTPS, but the new installer was never moved to HTTPS. This PR corrects the issue to allow the Yubikey integration to work out of the box.


s3inlc:
Thanks for the contribution!

PR #806, opened 2022-08-19
Added missing descriptions for user api commands

s3inlc:
Some descriptions were not added and were showing up as __<id>__ instead of a readable text.


zyronix (approved):
Looks good

Issue #805, opened 2022-08-19
Feature Request: Allow setting an agent to be CPU only from the agent side.


s3inlc:
Add a flag to the agent (e.g. --cpu) which makes the agent only querying CPUs and register on the server side as a CPU only agent directly.
This allows easier deployment of agents which should have two agents running (one for GPUs and one for CPUs) without having the GPUs listed on both agents on the server overview, and additionally not having to set to CPU only on the server side after registration.

zyronix:
Fixed in #807

PR #807, opened 2022-08-22
Allowing an agent to register as cpu agent directly

s3inlc:
This extension is part of the feature requested in issue #805.

It allows an agent to provide a flag denoting that it should be set as CPU agent directly.


zyronix (changes requested):
I have given this feature quite some though

I think we should not implement this. This allow you to set value on the server-side through the client. This can potentially introduce a security risk. Task can now be label as CPU only, if a node is not mark as CPU it currently has no access to to tasks. If you introduce this feature, you can potentially get access to tasks which you do not have access to normally.

I know cpu-only is not a security group or anything like that, but the point is more: either you trust clients to set values or you do not trust them. Imo it is bad to allow for change server values from a client and you should not allow this.

zyronix (approved):

PR #814, opened 2022-09-07
Initial POC for split user interface and new API

cc-nfi:
This is a first POC for a new split UI and backend API. We would like to use this PR to discuss if and which aspects would be good to keep and expand on, and what should be changed and improved. For this proposal we made use of Luracast Restler (https://restler5.luracast.com/) for the backend, and Angular (https://angular.io/) for the frontend.

The focus of the implementation is on the backend side. So the most important thing to review is everything under `src/api/v2`. The existing frontend and user/client APIs remain unchanged. We have added some basic API calls for hashlists, tasks and some other small things. Most of these are protected and require an API key to be passed. We also added DTO's for the communication, which are also validated up front by the framework we use. For testing, we made use of Behat, for which we added a couple of sample test cases. We also enabled Swagger documentation generation, which can also be used to manually test the API.

The UI code is an extremely simple POC, just to show the idea. With it, you can create hashlists and view/update tasks.

If you want to test this, you could check out this branch and run it inside the devcontainer. If all goes well, you should be able to:
- call into the new API, which resides at `:8080/api/v2/*`
- look into the Swagger documentation (and test the API) at `:8080/api/v2/explorer` (pass a generated access key in the upper right)
- view the basic UI at `:4200` (requires setting the access key in `configuration.ts`)
- run the Behat test suite by executing `../../vendor/bin/behat` inside directory `/var/www/html/ci/apiv2` in the Hashtopolis container

Note that there are still both implementation (e.g. clean setup code of tests, move Factory calls to Utils/Service, clean file importing), as wel as technical issues (e.g. caching in production mode, Swagger doc generation is iffy) to resolve. And perhaps other bugs (we would expand the tests and review everything in depth when we'd have an idea of way forward regarding implementation).

We are looking forward to your thoughts.

swagger
behat
ui


zyronix:
Continues in #832

PR #817, opened 2022-09-14
Add display of 'Use static chunking' settings in task details page

### LINKED ISSUE #816: Feature request: display static chunking choices on task details page
When creating a new task 'Use static chunking' options can be selected:
image

How-ever when viewing task details those settings are not shown, which is an inconvenience when debugging poorly running task. It would be nice if the 'task details' UI shows this.  Two examples:

a)
image

b) 
image

rixvet:
Proof-Of-Concept code for issue #816


src/templates/tasks/detail.template.html
```diff
@@ -234,6 +234,20 @@ <h2>Task details{{IF [[task.getUsePreprocessor()]]}} (Preprocessor task){{ENDIF}
           {{ENDIF}}
         </td>
       </tr>
+      <tr>
+        <td>Use static chunking:</td>
+        <td>
+          {{IF [[task.getStaticChunks()]] == 0}}
```
s3inlc, line 240: The values for the StaticChunks value are defined as constants in https://github.com/hashtopolis/server/blob/master/src/inc/defines/tasks.php#L40, so ideally these should be used (and in line 243 below as well). 
This can be used by using constants in the templating, e.g. `[[$DTaskStaticChunking::NORMAL]]`.

src/templates/tasks/detail.template.html
```diff
@@ -234,6 +234,20 @@ <h2>Task details{{IF [[task.getUsePreprocessor()]]}} (Preprocessor task){{ENDIF}
...
+            No
+          {{ELSE}}
+            {{IF [[task.getStaticChunks()]] == 1}}
```
s3inlc, line 243: see above

s3inlc (changes requested):

zyronix:
@rickvanderzwet is this one ready for another round of review?

rickvanderzwet:
@zyronix I addressed the suggested changes. It's ready for a next round of review. I already marked the comments 'resolved' a while ago, but I am guessing I also need mark it somewhere else as well?

s3inlc:
This is solved in the new frontend, the old frontend is deprecated.

PR #832, opened 2022-11-24
Initial POC for split user interface and new API

zyronix:
See #814 for bit of background info on this. This is a new approach for the development using Slim Framework instead of Restler.


zyronix:
Moved to #845 change from @rickvanderzwet namespace to hashtopolis namespace.

PR #848, opened 2023-01-20
Fix description of hash type 1430

zyronix:
As per discord, the description for hashmode 1430 is wrong:

image

image

@s3inlc how shall we deal with this? I can create an update call to compare the description of this hashtype, or shall we only fix this for new installs like this pull requests does?


s3inlc:
I think it would be safe to also check this in the update scripts. We could even only change it, if it is exactly the wrong value (so we can be sure it was not changed already)

s3inlc:
Oh and 1440 is wrong as well @zyronix

rickvanderzwet (approved):

rickvanderzwet (changes requested):
Revert testing 'Approve' workflow test

zyronix:
@s3inlc looks good, but because I created it; I cannot approve it. Please approve and merge.

s3inlc:
> @s3inlc looks good, but because I created it; I cannot approve it. Please approve and merge.

Thanks for the feedback, as I did some of the changes I wanted you to have a look at it. Will therefore approve and merge.

s3inlc (approved):

Issue #851, opened 2023-02-02
Increase Hash.hash to mediumtext


Soren17:
When uploading a hash created by luks2hashcat.py, htp blanks out/crashes. The reason is that the hash is too big for the Hash table. I fixed this by running "alter table Hash modify hash mediumtext".
I suggest to change the table Hash, by changing the column hash from TEXT to MEDIUMTEXT.

zyronix:
Thanks, the reason for it being text is performance. If you have large hashlist changing it to medium can cause some performance issues there.

zyronix:
Duplicated from #841

Issue #841, opened 2022-12-16
Hashes over 65535 in length triggers database error


rickvanderzwet:
The  `./hashcat/tools/luks2hashcat.py` tooling (can) produce hashes which are over 65535 in length. Which are to be used together with LUKS hash algoritm 29521 (new in hashcat 6.2.6). When trying to insert into this into hashtopolis it triggers an database error:

```
PHP Fatal error:  Uncaught PDOException: 
SQLSTATE[42000]: Syntax error or access violation: 1074 Column length too big for column 'hash' (max = 65535); use BLOB or TEXT instead in /var/www/hashtopolis/inc/Util.class.php:1301\nStack trace:\n#0
/var/www/hashtopolis/inc/Util.class.php(1301): PDO->query('ALTER TABLE
Has...')\n#1 /var/www/hashtopolis/inc/utils/ConfigUtils.class.php(88): 
Util::setMaxHashLength(264000)\n#2
/var/www/hashtopolis/inc/handlers/ConfigHandler.class.php(13): 
ConfigUtils::updateConfig(Array)\n#3
/var/www/hashtopolis/config.php(22): 
ConfigHandler->handle('updateConfig')\n#4 {main}\n  thrown in
/var/www/hashtopolis/inc/Util.class.php on line 1301
```

Unable to insert long hashes has been reported a while ago in #184 and #324 how-ever both tickets are closed, allowing me to make a fresh one, since IMHO the issue still persist.

One suggestion would be to use a different way of storing large hashes in the database e.g.:
```Diff
diff --git a/src/install/hashtopolis.sql b/src/install/hashtopolis.sql
index 1b357b35..c3faff58 100644
--- a/src/install/hashtopolis.sql
+++ b/src/install/hashtopolis.sql
@@ -237,7 +237,7 @@ CREATE TABLE `FileDelete` (
 CREATE TABLE `Hash` (
   `hashId`      INT(11)      NOT NULL,
   `hashlistId`  INT(11)      NOT NULL,
-  `hash`        TEXT         NOT NULL,
+  `hash`        MEDIUMTEXT   NOT NULL,
   `salt`        VARCHAR(256) DEFAULT NULL,
   `plaintext`   VARCHAR(256) DEFAULT NULL,
   `timeCracked` BIGINT       DEFAULT NULL,
```

How-ever I am not sure whether this has unwanted side-effects.

zyronix:
Note for self: Double check how indexing of mysql is dealing with this.

PR #867, opened 2023-03-15
change type of Hash.hash to MEDIUMTEXT instead of TEXT to allow longer hashes

s3inlc:
closes issue #851 
closes issue #841


src/install/updates/update_v0.13.x_v0.x.x.php
```diff
@@ -10,4 +12,9 @@
 require_once(dirname(__FILE__) . "/../../inc/defines/config.php");
 require_once(dirname(__FILE__) . "/../../inc/defines/log.php");
 
+if (!isset($PRESENT["v0.13.x_hash_length"])) {
+  Factory::getAgentFactory()->getDB()->query("ALTER TABLE `Hash` MODIFY `hash` MEDIUMTEXT NOT NULL;");
```
zyronix, line 16: What if the database type is set to longtext? This change would modify that and thus cause issue when upgrading. Maybe add a check to this to check if the type is TEXT.

zyronix (approved):

zyronix:
@s3inlc I have fixed my own suggestion to double check that the column type is TEXT or smaller. I know that some people ahve already increased the size to LONGTEXT. This should prevent that from erroring or reverting back to MEDIUMTEXT. Could you verify my changes?

s3inlc:
Thanks, looks good. You're right, we should only increase if it's smaller.

Issue #865, opened 2023-03-13
Check for import flag when removing preconfigured task from supertask


s3inlc:
When a preconfigured task is removed from a supertask it normally is just a remove of the link as the preconfigured task still could be used. In the case the preconfigured task is from an import it should be deleted as it was specifically created for the supertask. This check currently is not done in SupertaskUtils `removePretaskFromSupertask()`, so the preconfigured task ends up not being shown in the list but not being connected to any supertask.

PR #868, opened 2023-03-15
when a pretask is removed from a supertask it should be deleted in the case it was created by an import

s3inlc:
closes issue #865

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -2,7 +2,7 @@
 
 ## Bugfixes
 
-- When a preconfigured task is removed from a supertask and its originating from an import, it should be deleted (issue #865).
+- When deleting a supertask that was created from an import, pretasks that were removed from this supertask should also be deleted (issue #865).
 
 # v0.12.0 -> v0.13.0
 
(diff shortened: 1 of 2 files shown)


doc/changelog.md
```diff
@@ -1,3 +1,9 @@
+# v0.13.0 -> v0.x.x
+
+## Bugfixes
+
+- When a preconfigured task is removed from a supertask and its originating from an import, it should be deleted (issue #865).
```
zyronix, line 5: The description is not clear enough.

Suggestion:

`When deleting a supertask that was created from an import, pretasks that were removed from this supertask should also be deleted.`

zyronix (changes requested):

zyronix:
> I cannot confirm this issue. When in master I create a supertask with masks through the import function and delete the supertasks all my pretasks get deleted.

Nvm, I found the correct flow to reproduce this. Code works, but I have a small text suggestion for the changelog.

doc/changelog.md
s3inlc: The text is now changed to your suggestion.

PR #893, opened 2023-03-31
Removing bugged check from HashlistHandler

zyronix:
In the HashlistHandler there is a weird line. It is checking if the secret value is set in the post when checking for the hexsalted value:

```
(isset($_POST["hexsalted"]) && (isset($_POST["secret"]) && intval($_POST["secret"]) == 1) && intval($_POST["hexsalted"]) == 1) ? true : false,
```

This is added since version 0.7.0 already, and this causes the hexsalted value to not work correctly. As it could not be set when creating hashlist through the UI. This pull requests fixes that.

I have spent some time for adding a test for this aswell, but because the HashlistHandler requires a lot of UI elements and in the end fire a `die()` call, I could not create a test for it.

### DIFF
--- a/doc/changelog.md
+++ b/doc/changelog.md
@@ -1,3 +1,7 @@
+# v0.13.1 -> v0.x.x
+## Bugfixes
+- Setting 'Salt is in hex' during Hashlist creation will not set the --hex-salt flag (#892)
+
 # v0.13.0 -> v0.13.1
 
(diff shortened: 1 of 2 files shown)


s3inlc (approved):

s3inlc:
Good find! I guess this originated from a copy-paste mistake.

Issue #914, opened 2023-06-09
Release 0.14.0 cleanup


zyronix:
With the release of 0.14.0 some stuff need some cleaning. Due to some time constrains @s3inlc and me were not able to do everything. This issue is there just to check what we still need to do.

 - [ ] Fix github actions
 - [ ] Fix running test locally
 - [ ] Clean branches
 - [ ] Docker description text for repo and organisation.
 - [ ] Test/fix  devcontainer, implement environment variable
 - [ ] Update install documentation
 - [ ] Marketing -> discord, blogpost, twitter
 - [ ] Celebrate

s3inlc:
Tests are fixed with #916

PR #935, opened 2023-06-20
devcontainer fixes for working with 'production' docker creation #914 

rixvet:
Fix dev container:
 - Allow building next to 'production' docker container (https://github.com/hashtopolis/server/wiki/Installation#docker-setup).
 - Enhance syntax highlighing by adding php composer filles in visible space.
 - Fix python test packages not able to install.
 - Enable APIv2 by default.
 - Fix .env template to work with changed DOCKER variables for working behind corporate firewalls.

Fixes are part of #914


zyronix:
Pushed some small changes to the container. Removed some logic from vscode specific file to the generic docker file so the contaiener can also be used without vscode.

Issue #936, opened 2023-06-22
[BUG] Failed to login after recreating the devcontainer


zyronix:
1. Create devcontainer
2. Close and delete devcontainer
3. Start the devcontainer again
4. Cannot login anymore

Why? Because the database will keep using its old volume, but the backend will generated a new pepper.

zyronix:
Pull request ready, waiting for review

PR #937, opened 2023-06-22
Fixes 936

zyronix:
Adding volume to devcontainer for making config.json persist

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -22,6 +22,7 @@ services:
       # This is where VS Code should expect to find your project's source code
       # and the value of "workspaceFolder" in .devcontainer/devcontainer.json
       - ..:/var/www/html
+      - hashtopolis-server-dev:/usr/local/share/hashtopolis:Z
     networks:
       - hashtopolis_dev
   hashtopolis-db-dev:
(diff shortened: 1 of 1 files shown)


.devcontainer/docker-compose.yml
```diff
@@ -22,6 +22,7 @@ services:
       # This is where VS Code should expect to find your project's source code
       # and the value of "workspaceFolder" in .devcontainer/devcontainer.json
       - ..:/var/www/html
+      - hashtopolis-server-dev:/usr/local/share/hashtopolis:Z
```
rixvet, line 25: The ':Z' only seems relevant on systems running selinux policies. (See: https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label). Looks optional (jet harmless) here.

rixvet (approved):
This fixes behavior #936 with two side-notes: 

 -  The pepper file (on persistent storage) is only created when the database is not yet populated. Thus an re-run against an existing dev-container database does not create the new persistent config files. This edge case how-ever only occurs on 'transition'. Extra 'Close and delete devcontainer-db as step 2b' fixes it.

 - Rebuilding the dev-container while keeping /usr/local/share/hashtopolis makes the dev container setup behave more consistent with the production container setup. Since all underlying files are also kept:
/usr/local/share/hashtopolis/import
/usr/local/share/hashtopolis/config
/usr/local/share/hashtopolis/config/config.json
/usr/local/share/hashtopolis/files
/usr/local/share/hashtopolis/log

Issue #947, opened 2023-07-12
[DOC] User-API documentation misses `deleteAgent` entry


robinduerhager:
Hi there,

i just noticed that the `deleteAgent` API route of the `user-api` is missing and wanted to let you know :).
Should i create a PR for this?

PR #953, opened 2023-07-24
[DOC] Added missing `deleteAgent` entry to documentation of the user-api

robinduerhager:
As mentioned in #947 , this PR adds an entry of the implemented, but documentation wise missing `deleteAgent` functionality to the documentation of the user-api.

In the changelog i didn't find any records of `added documentation for xyz`. Should something like this be added into the changelog as well?


zyronix:
@s3inlc Could you review this one?

s3inlc:
Thanks for submitting this PR :)

I think there is no need for having such changes explicitly mentioned in the changelog.

PR #956, opened 2023-07-25
Idea: harden innitialisation

rixvet:
@zyronix I would like to add some pre-flight checking to avoid errors when files cannot be written. Spotting this errors early in the process allows for some robust error messages during start/init. 

Do you see any downsides in adding them?


zyronix:
Not something I can think off. Looks good.

Issue #960, opened 2023-08-16
Adding volume mount for custom apache2 config


robinduerhager:
It would be nice if we could have a volume in the docker installation, in which we can provide our own `000-default.conf` apache2 config in case of some individualized webserver settings.

ObsidianOracle:
Mounts can always be done within docker. If you still have any doubts, please reach out.

PR #961, opened 2023-08-16
WIP: Add SSL integration to apache2 configuration

robinduerhager:
Hey,

i extended the `000-default.conf` with a simple ssl configuration.
Currently you would have to add a `cert.pem` and `cert-key.pem` to `.devcontainer/ssl/`.

You can do this by running [mkcert](https://github.com/FiloSottile/mkcert) like this:

```
mkcert -install
mkcert localhost 127.0.0.1 ::1
```

Is the apache config in `.devcontainer/sites-enabled/` meant to be configured by the user? If so, then i guess #960 is fixed as well? :)

It does run without problems in the devcontainer.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -18,6 +18,7 @@ services:
       - hashtopolis-db-dev
     ports:
       - "8080:80"
+      - "9443:443"
     volumes:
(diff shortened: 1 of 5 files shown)


robinduerhager:
PR was based on master, will do a new one, based on newest dev changes.

PR #964, opened 2023-08-23
Fix incorrect parameter checked

### LINKED ISSUE #963: [BUG] Pretask attackCmd parameter not checked for length on create
``Attack command is too long (max 256 characters)!`` does not seems to get triggered

rixvet:
Fixes #963


zyronix (approved):

zyronix (changes requested):
Can you add somelines to the changelog about this?

zyronix (approved):
Thanks!

Issue #969, opened 2023-08-24
[BUG] SupertaskUtils::runSupertask fails if pretask CrackerBinaryType to provided $crackerBinaryType


rixvet:
**Current Behavior**
Calling SupertaskUtils::runSupertask with pretask where CrackerBinaryType is not equal to $crackerId BinaryType will cause an error "Object of class DBA\\CrackerBinary could not be converted to string" 

**Expected Behavior**
function exectutes with error

Background
There are conditional statement [if ($cracker->getCrackerBinaryTypeId() != $pretask->getCrackerBinaryTypeId()) {](https://github.com/hashtopolis/server/blob/a54c07cd76ec1feda0e8a8abb78dfda5ea77103d/src/inc/utils/SupertaskUtils.class.php#L280)
which triggers an object returned instead of an Id.

Additional information
a) PoC fix
```
diff --git a/src/inc/utils/SupertaskUtils.class.php b/src/inc/utils/SupertaskUtils.class.php
index 004fb1cb..b4276577 100644
--- a/src/inc/utils/SupertaskUtils.class.php
+++ b/src/inc/utils/SupertaskUtils.class.php
@@ -278,7 +278,7 @@ class SupertaskUtils {
     foreach ($pretasks as $pretask) {
       $crackerBinaryId = $cracker->getId();
       if ($cracker->getCrackerBinaryTypeId() != $pretask->getCrackerBinaryTypeId()) {
-        $crackerBinaryId = CrackerBinaryUtils::getNewestVersion($pretask->getCrackerBinaryTypeId());
+        $crackerBinaryId = CrackerBinaryUtils::getNewestVersion($pretask->getCrackerBinaryTypeId())->getId();
       }

       $task = new Task(
```

PR #971, opened 2023-08-28
Fix create supertask fails when crackerType != pretask.crackerType 

rixvet:
Instead of returning an object, it  should return an ID instead.

Fixes: #969


zyronix:
Looks good, could you please add something to the changelog?

This condition is only triggered when you for example create a supertask bet select the cracktype incorrectly. Eg, you want to run a task with the generic cracker, but the prestask inside the supertask were created with hashcat.

zyronix (changes requested):
See above, update changelog

zyronix (approved):

Issue #989, opened 2023-09-05
[BUG]: Settings priority in #/tasks/show-tasks for normal task does not apply to taskwrapper


zyronix:
### Version Information

master

### Description

When I set the priority of a normal task (not a runningSuperTask). The priority is not applied on the TaskWrapper database object, and thus the priority is not implemented.

When patching priority and maxagent on a Task object, also reverse patch the taskwrapper object if it is not a supertask.

rixvet:
This almost feels something which is a necessary evil to implement. It breaks the basic concepts of the APIv2/ui section modelling since this will causes changes at one model be effected to other models 'magically', which should be an helper function, how-ever since going forward we want to hide the 'TaskWrapper' object (again see #906) so let's go for it :innocent:

zyronix:
Another solution could be to make the task view about viewing taskwrappers. Once clicked you'll view the tasks inside the taskwrapper; even when only one task is available. The only behavior is that when tasktype is task and not supertask, the priority and maxagents cannot be patched through the api. As they can only bet set on the taskwrapper. What you think about this idea?

rixvet:
> Another solution could be to make the task view about viewing taskwrappers. Once clicked you'll view the tasks inside the taskwrapper; even when only one task is available. The only behavior is that when tasktype is task and not supertask, the priority and maxagents cannot be patched through the api. As they can only bet set on the taskwrapper. What you think about this idea?

I think it is rather confusing, both from a design point-of-view as from user point-of-view, since this means we updating/creation behavior based on conditionals within the object. Maybe hiding the taskWrapper completely and deal with an new endpoint for 'group of tasks' makes it easier to understand. Let's think of it in the context of #991.

PR #990, opened 2023-09-05
Fix update priority on task does not apply to taskwrapper

rixvet:
The task object priority is servering asl  'dummy' for  normal tasks.

Related taskwrapper object should be updated as well,
since the 'real' priority scheduling is done using the
Taskwrapper object.

Fixes: #989


zyronix:
@rixvet Kind of mist this one, is this one still relevant? Does it make sense to merge it into master or can we just push it to dev?

zyronix (approved):

PR #1026, opened 2023-11-19
Added Vast.AI Integration

deadjakk:
You probably are already aware of [vast.ai](https://vast.ai), it's uber for GPUs.   
This PR adds a few more pages to hashtopolis to facilitate 1-2 click en-masse deployment
of agents by leveraging vast.ai's API.  
The goal was to add the ability to deploy and destroy complete cracking
infrastructure in as few clicks as possible.

I am also happy to maintain this portion of code if needed, and will plan to add the same functionality to the node version of hashtopolis once that is in a more mature state.  

One caveat: vast.ai instances are not all created equal, therefore some still throw 'weird' errors from time to time but using the default search parameters I have provided in the code I have had pretty good luck; about an 80%-90% success rate on average, YMMV.

It adds:
- Vast.ai instance search/rental page with a multi-purchase option
- Vast.ai instance management page using live values from vast.ai and an auto-refresh function
- Auto deployment upon renting using a [docker image I published](https://hub.docker.com/repository/docker/deadjakk/hcat-vast-dev/general) (Dockerfile provided below)
- New config values under config > Server > Vast.ai section to support this

It is missing:
1. docs - I am waiting to see if this is going to be accepted before bothering with this
2. auto agent deletion when an instance is released/destroyed from vast.ai - to do this properly, I will need to add a 'tag' flag to the hashtopolis agent and server api so that I can be certain about which instance is which
3. custom cert package downloads to allow for unsigned ssl connections  
4. update file addition to add the new configuration values, something like the code block below 

```
$config = new Config(null, 8, DConfig::VAST_AI_API_KEY, '');
Factory::getConfigFactory()->save($config);
$config = new Config(null, 8, DConfig::VAST_IMAGE_URL, '');
Factory::getConfigFactory()->save($config);
$config = new Config(null, 8, DConfig::VAST_IMAGE_LOGIN, '');
Factory::getConfigFactory()->save($config);
$config = new Config(null, 8, DConfig::VAST_HASHTOPOLIS_BASE_URL, '');
Factory::getConfigFactory()->save($config);
```

Below is the Dockerfile so you can publish it via your own hub.docker.com account and have control over it.   
# Dockerfile
```
FROM nvidia/cuda:12.2.2-devel-ubuntu22.04
# Example of successful usage command
# YOU MUST HAVE VALID SSL CERTS IF USING AN HTTPS SERVER !!!!
#
# to run independent of hashtopolis:
# sudo docker build . -t hcat
# sudo docker run -it -e HCATURL="https://localhost:9090" -e VOUCHER="GF30HJEx" hcat

RUN apt update && \
   apt install -y curl zip python3 python3-psutil python3-requests pciutils pocl-opencl-icd

RUN mkdir -p /root/hcat

WORKDIR /root/hcat

RUN echo 'curl -k "$HCATURL/agents.php?download=1" -o agent.zip' > run.sh
RUN echo 'unzip agent.zip' >> run.sh
RUN echo 'python3 . --voucher $VOUCHER --url "$HCATURL/api/server.php"' >> run.sh
RUN chmod +x run.sh

## This will not get run by vast.ai, so this really only exists for use outside of vast.ai
CMD curl -k "$HCATURL/agents.php?download=1" -o agent.zip && \
    unzip agent.zip && \
    python3 . --voucher $VOUCHER --url "$HCATURL/api/server.php"
```

# Images of use  
Searching for instances
search-instances
Renting instance
successful-rent
Rented instances
rented-instances
Agents 'received'
instace-agents

### DIFF
--- a/src/inc/defines/config.php
+++ b/src/inc/defines/config.php
@@ -108,6 +108,12 @@ class DConfig {
   const NOTIFICATIONS_PROXY_PORT   = "notificationsProxyPort";
   const NOTIFICATIONS_PROXY_TYPE   = "notificationsProxyType";
   
+  // Section: VastAI
+  const VAST_AI_API_KEY = "vastAiApiKey";
+  const VAST_IMAGE_URL = "vastImageUrl";
+  const VAST_IMAGE_LOGIN = "vastImageLogin";
+  const VAST_HASHTOPOLIS_BASE_URL = "vastHashtopolisBaseUrl";
+
   static function getConstants() {
     try {
       $oClass = new ReflectionClass(__CLASS__);
@@ -272,6 +278,14 @@ class DConfig {
         return DConfigType::TICKBOX;
       case DConfig::HC_ERROR_IGNORE:
         return DConfigType::STRING_INPUT;
+      case DConfig::VAST_AI_API_KEY:
+        return DConfigType::STRING_INPUT;
+      case DConfig::VAST_IMAGE_URL:
+        return DConfigType::STRING_INPUT;
+      case DConfig::VAST_IMAGE_LOGIN:
+        return DConfigType::STRING_INPUT;
+      case DConfig::VAST_HASHTOPOLIS_BASE_URL:
+        return DConfigType::STRING_INPUT;
     }
     return DConfigType::STRING_INPUT;
   }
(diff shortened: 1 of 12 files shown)


zyronix:
Thanks for your contribution! Sorry for the late response. There have been multiple discussion on simlar topic before: should we add device management to Hashtopolis (Like wake-on lan support). If so, in which form?

The problem with allowing it, why allow vast.ai now? And how about bare metal, how about google, aws etc.

My personal take on this: Hashtopolis is for distributing hashcat workload not for node management. I do however understand the need for integration with some kind of node manamgent; but I don't think that adding all platform seperatly is the right way for this.

deadjakk:
Have there been any thoughts on adding a plugin interface for these sorts
of things?

ovcrash:
This is very interesting. In my opinion, this is not node management but on-demand workload management. I understand that the underlying aspect of it, is managing the nodes.
It should clearly be integrated into Azure and AWS, since those are big players.

deadjakk:
> This is very interesting. In my opinion, this is not node management but on-demand workload management. I understand that the underlying aspect of it, is managing the nodes. It should clearly be integrated into Azure and AWS, since those are big players.

I went with vast because it was the cheapest I could find, but I am open to adding similar AWS/Azure support. Would it be merged in that case?

zerlokker (approved):

PR #1029, opened 2023-12-21
NotificationSlack: Set the content-type header to Application/json. 

julianjm:
This makes it compatible with Google Chat webhooks.

https://developers.google.com/chat/how-tos/webhooks#write_the_webhook_script


maede97:
This is also required for Mattermost. We would highly appreciate if this can be fixed soon.

s3inlc:
I incorporated the changes into a new PR due to the changes and resulting conflicts: #2372

PR #1043, opened 2024-02-21
Implemented helper for password reset

s3inlc:
Helper added for non-authenticated requests for a reset of a users password.

This function requires emails to be sent, therefore the email docker configuration is included as well (which for example would also be used by email notifications).

As a simple sendmail utility, ssmtp is used whose configuration is now included in docker-compose and the example env as well.


jessevz (changes requested):

s3inlc:
The process to get the email support is working that ssmtp is installed in the docker and if email support is wished, the config file in /etc/ssmtp/ssmtp.conf can just be mounted into the docker via volumes

jessevz (approved):

PR #1062, opened 2024-04-22
Adding loops to scan through lines to support importing hashes longer then 1024 bytes

zyronix:
Added couple of loops that allows to import hashes that are larger then 1024 bytes.


zyronix:
Note, that if you offer a very large file. These code changes will loop and keep reading the file until the memory limit of the process is hit.

s3inlc (approved):

PR #1076, opened 2024-09-13
Ipv6 support in buildServerUrl()

### LINKED ISSUE #725: [FEATURE] IPv6 support
Latest release `0.12.0` (and early) not support work via IPv6 connection. For example, start page says:
```
To add new agents provide them with a valid voucher and download a client.
Used vouchers are automatically deleted to prevent double spending.
If you are asked to provide the API url on the client to connect to, you need to enter the following:
http://[201/api/server.php
```
`http://[201/api/server.php` instead `http://[201:a373:f940:2421:b2f2:6a91:86b7:5628]/api/server.php`

Also install procedure via IPv6 connection has everytime stopped at connection to db step. Via localhost works correctly.

Thanks.

jessevz:
fixes #725 

The current code didn't work in case of an ipv6 address. This is because the host name got parsed by using strpos($hostname, ":"). this goes perfectly fine for an ipv4 address. But when you have an ipv6 address that already contains multiple ":" characters this goes wrong. For example http://[201:a373:f940:2421:b2f2:6a91:86b7:5628]/api/server.php would become http://[201/api/server.php. I fixed it by using strrpos when the server is not at port 80 or 433. This way the last occurrence of ":" will be used which splits the hostname of the port.

### DIFF
--- a/ci/HashtopolisTest.class.php
+++ b/ci/HashtopolisTest.class.php
@@ -34,7 +34,8 @@ abstract class HashtopolisTest {
     "0.13.0" => "c7036800be5b2e21542df0fa9bd4d19ebf7ecdf3",
     "0.13.1" => "fc4b5226c1d36dc0197f4d17d216298972055c09",
     "0.14.0" => "0af7193310c9c1f37a274578e159ab3ae0a6caad",
-    "0.14.1" => "375f2ce022c4b3e0780abf9dcca1e6af8e966c1a"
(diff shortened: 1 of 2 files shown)


jessevz:
ah yes i see its because i forked the main branch and i made a pull request for the dev branch and apparently the main branch has more commits than the dev branch. I could revert the commits but i dont know if it really matters because with this pull request the dev branch will be up to date again with the main branch.

gluafamichl:
Can we check an merge this PR or do you want to revert the changes made for HashLists?

gluafamichl (approved):

PR #1102, opened 2024-10-10
Fix for mismatch # of parameters with importSupertask in the User API

matt-moses:
Fix for mismatch # of parameters with importSupertask in the User API. I'm not super familiar with the codebase still so please review and let me know if there is a more correct fix.

https://github.com/hashtopolis/server/blob/master/src/inc/utils/SupertaskUtils.class.php#L362 and
https://github.com/hashtopolis/server/blob/master/src/inc/user-api/UserAPISupertask.class.php#L114C21-L121C41

Also, some documentation updates. What is the process for rebuilding the PDF? Can someone help with that please?

Thanks for a great tool!

### DIFF
--- a/doc/user-api/sections/supertask.tex
+++ b/doc/user-api/sections/supertask.tex
@@ -9,6 +9,7 @@
 			  "section": "supertask",
 			  "request": "importSupertask",
 			  "name": "Mask Supertask",
+			  "maxAgents": 10,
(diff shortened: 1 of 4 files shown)


s3inlc:
As the user-api is deprecated, there will be no further changes made to the docs which will get removed.

Issue #1101, opened 2024-10-10
[BUG]: Cracked hashes not appearing in GUI or API


yungcero:
### Version Information

0.14.2

### Hashcat

6.2.6

### Description

When accessing the "cracked" section of hashtopolis, whether through the GUI or API (getCracked request) will return empty values even though hashes are verified to be cracked and can be seen when clicked on the link. The hashlists being cracked are composed of one hash.

The bottom shows what happens when clicking on Cracks:

<img width="1530" alt="Screenshot 2024-10-10 at 10 52 47 AM" src="https://github.com/user-attachments/assets/143da158-6c12-4f92-8958-6afd7bb6b384">

However hashlists show it is cracked:

<img width="1031" alt="Screenshot 2024-10-10 at 10 53 29 AM" src="https://github.com/user-attachments/assets/20ccd1a3-b96a-4e76-bac0-bca2fa49aea0">
<img width="1569" alt="Screenshot 2024-10-10 at 10 53 08 AM" src="https://github.com/user-attachments/assets/9c69529c-be73-4cf0-9d60-0ce446ce28a7">

And with a publicly available hash, shows the plaintext:

<img width="1528" alt="Screenshot 2024-10-10 at 10 53 43 AM" src="https://github.com/user-attachments/assets/4527c386-c60a-485a-a746-ed70ec9c20fd">

Cracked hashes can only be accessible when clicking on the link chain. When using the api or the cracks tab nothing shows for the hashlist as it being cracked.  It is using runtime benchmark and this is the command: `#HL# rockyou.txt -d 2 --hwmon-disable`. I can also produce the hash for testing if needed as it is a publicly accessible one on Github used for CTFs

yungcero:
I also get this error with my agent after i crack a hash. Not sure if its related but want to throw it in here just in case

```
Client is up-to-date!
Got cracker binary type hashcat
Keyspace got accepted!
Benchmark task...
Server accepted benchmark!
Start chunk...
Progress: 41.11% Speed:   4.01MH/s Cracks: 1 Accepted: 1 Skips: 0 Zaps: 0
finished chunk
Client is up-to-date!
Getting of file failed: {'action': 'getFile', 'response': 'ERROR', 'message': 'Client is not assigned to this task!'}
```

yungcero:
Going to fork the repo and push up some changes, think i have a bandaid fix for at least the api side

jessevz:
[...]

PR #1108, opened 2024-10-15
FIX: fix issue where hashes that are in binary format do not get thei…

yungcero:
_…r_ cracked plaintext returned through the api. 

Is able to return information by including in the HashBinary factory in the GetCracked function. Helps address issue #1101

### DIFF
--- a/src/inc/utils/HashlistUtils.class.php
+++ b/src/inc/utils/HashlistUtils.class.php
@@ -1130,35 +1130,23 @@ class HashlistUtils {
     if (!AccessUtils::userCanAccessHashlists($lists, $user)) {
       throw new HTException("No access to the hashlists!");
     }
-    
     $hashlistIds = Util::arrayOfIds($lists);
     $qF1 = new ContainFilter(Hash::HASHLIST_ID, $hashlistIds);
     $qF2 = new QueryFilter(Hash::IS_CRACKED, 1, "=");
     $hashes = [];
-    $format = Factory::getHashlistFactory()->get($hashlistId);
-    if ($format->getFormat() != 0) {
-      $entries_binary = Factory::getHashBinaryFactory()->filter([Factory::FILTER => [$qF1, $qF2]]);
-      foreach ($entries_binary as $entry) {
-        $arr = [
-          "hash" => $entry->getHash(),
-          "plain" => $entry->getPlaintext(),
-          "crackpos" => $entry->getCrackPos()
-        ];
-        $hashes[] = $arr;
+    $isBinary = $hashlist->getFormat() != 0;
+    $factory = $isBinary ? Factory::getHashBinaryFactory() : Factory::getHashFactory();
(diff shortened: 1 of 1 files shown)


src/inc/utils/HashlistUtils.class.php
```diff
@@ -1147,6 +1147,15 @@ public static function getCrackedHashes($hashlistId, $user) {
       }
       $hashes[] = $arr;
     }
+    $entries_binary = Factory::getHashBinaryFactory()->filter([Factory::FILTER => [$qF1, $qF2]]);
```
jessevz, line 1150: Instead of doing 2 queries, it is better to check in the hashlist whether it is a binary hashlist or a normal text hashlist. What you could do is the following:

if ($hashlist->getFormat === 0) {
    $entries = Factory::getHashFactory()->filter([Factory::FILTER => [$qF1, $qF2]]);
} else {
    $entries = Factory::getBinaryHashFactory()->filter([Factory::FILTER => [$qF1, $qF2]]);
}

And then you could build the hashes according to their type. In the HashlistUtils line 700 to line 714 you can see how hashes of different types are built, you can use that as example.

jessevz (commented):
Thank you for your contribution! I’ve reviewed the code, and it looks solid overall. I appreciate the effort you’ve put in.

There is, however, one small issue I noticed: before performing the query, it would be ideal to first check what type of hash list is being used. This way only 1 query needs to be performed and based on the type you can then build the correct format of the hash. If you have any questions or need help to implement this please let me know.

Thanks again for your work!

src/inc/utils/HashlistUtils.class.php
yungcero: Awesome! Thanks for pointing that out. I have implemented the recommended change

jessevz:
Thanks for then changes, this looks better, I see some very minor issues. First of in line 1138 you retrieve the hashlist from the database while you already have access to it (line 1128). Also there is a little bit of code duplication, namely the way you loop is the same in both cases, you could solve that in something similar as the following way:

$factory = Factory::getHashFactory();
$format = $hashlist->getFormat();
if ($format > 0) {
    $factory = Factory::getHashBinaryFactory();
}

$entries = factory->filter([Factory::FILTER => [$qF1, $qF2]]);
foreach ($entries as $entry) {
$arr = [
  "hash" => $entry->getHash(),
  "plain" => $entry->getPlaintext(),
  "crackpos" => $entry->getCrackPos()
];
 
if ($hashlist->isSalted() != 0) {
        $arr["hash"] .= $hashlist->getSaltSeparator() . $entry->getSalt();
 }
$hashes[] = $arr;
}

Another solution could be to make a function for it so that we dont have a duplicate loop inside the code. But it looks good! When i have some more time the following weeks i will test it and merge it if there are no further issues

yungcero:
Awesome, ill work up those changes you mentioned and push up soon

yungcero:
Added updates! @jessevz

jessevz (approved):

PR #1116, opened 2024-10-28
Bug/1112 bug encoding error in the database or api

gluafamichl:
- add html_entity_decode for strings to db2json function
- remove duplicate htmlentities for Hashlists, Superhashlists and Tasks

### DIFF
--- a/src/inc/apiv2/common/AbstractBaseAPI.class.php
+++ b/src/inc/apiv2/common/AbstractBaseAPI.class.php
@@ -398,7 +398,7 @@ abstract class AbstractBaseAPI
       $obj = array_map('intval', preg_split("/,/", $val, -1, PREG_SPLIT_NO_EMPTY));
     } elseif ($feature['type'] == 'dict' && $feature['subtype'] = 'bool') {
       $obj = unserialize($val);
-    } elseif (str_starts_with($feature['type'], 'str')) {
+    } elseif (str_starts_with($feature['type'], 'str') && $val !== null) {
       $obj = html_entity_decode($val, ENT_COMPAT, "UTF-8");
     }
     else {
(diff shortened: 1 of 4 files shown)


src/inc/apiv2/common/AbstractBaseAPI.class.php
```diff
@@ -398,7 +398,10 @@ protected static function db2json(array $feature, mixed $val): mixed
       $obj = array_map('intval', preg_split("/,/", $val, -1, PREG_SPLIT_NO_EMPTY));
     } elseif ($feature['type'] == 'dict' && $feature['subtype'] = 'bool') {
       $obj = unserialize($val);
-    } else {
+    } elseif (str_starts_with($feature['type'], 'str')) {
```
jessevz, line 401: Currently 1 of the test fails, I think its because html_entity_decode will return an empty string when you input null. So now it fails when a task color is set to null because the test expects a null value but it gets an empty string. To solve it, I think inside this if statement you can just check whether $val != null and I think then it will pass the test. The elseif statement will then become:

} elseif (str_starts_with($feauture['type'], 'str') && $val !== null)

gluafamichl: Ah, thanks, changed the code like you proposed.

gluafamichl: Just one question, do you know how to run the tests locally before doing a commit or PR?

jessevz: Yes, in order to test the new api locally you can cd to the correct folder: "cd ~/src/hashtopolis/server/ci/apiv2" and then run the following command: "pytest --exitfirst --last-failed"

PR #1123, opened 2024-11-09
raise limits for upload files

satandyh:
Found that You cannot add file more than 8Mb (around). If you want to upload big hashtable/wordlist/etc file - it's blocker for You. So I just raised that limits to 256Gb.


bultodepapas:
please!!!!

PR #1126, opened 2024-11-12
Fixed preprocessor skip command bug.

gochujang-c:
Fixes small bug which would revert the skip command to the limit command when first creating through the UI.


zyronix (approved):
Please add a line to the changelog.md

jessevz:
Currently the pipeline is failing because of #1133. I will look into this is as quickly as possible and then merge this pull request.

PR #1127, opened 2024-11-12
Use utf8mb4 as default encoding

gochujang-c:
utf8 defaults to utf8mb3, which can only handle a subset of Unicode. An example which fails is:
```
└─>>> echo -n 🍌 | xxd
00000000: f09f 8d8c                                ....
└─>>> echo -n 🍌 | md5sum
806a7d7522baf4c10d2b45949075b382  -
```
If you create a hashlist with this md5 hash and try to import the solution as a pre-crack, you get the following error:
```
[12-Nov-2024 16:48:18 UTC] PHP Fatal error:  Uncaught PDOException: SQLSTATE[HY000]: General error: 1366 Incorrect string value: '\xF0\x9F\x8D\x8C' for column 'plaintext' at row 1 in /var/www/html/src/dba/AbstractModelFactory.class.php:244
Stack trace:
#0 /var/www/html/src/dba/AbstractModelFactory.class.php(244): PDOStatement->execute(Array)
#1 /var/www/html/src/inc/utils/HashlistUtils.class.php(478): DBA\AbstractModelFactory->mset(Object(DBA\Hash), Array)
#2 /var/www/html/src/inc/handlers/HashlistHandler.class.php(52): HashlistUtils::processZap('5', ':', 'paste', Array, Array, Object(DBA\User))
#3 /var/www/html/src/hashlists.php(31): HashlistHandler->handle('processZap')
#4 {main}
  thrown in /var/www/html/src/dba/AbstractModelFactory.class.php on line 244
```
Seems that clients since version 8.0 use this as the default, so removing it from the connection string would then work automatically, but perhaps keep it explicit/other use case?

For changing it server side (if necessary), you could add something like the following to `/etc/my.cnf`:
```
[client]
default-character-set = utf8mb4
 
[mysql]
default-character-set = utf8mb4
 
[mysqld]
character-set-server = utf8mb4
collation-server = utf8mb4_0900_ai_ci
``` 
and restart the server.


src/dba/AbstractModelFactory.class.php
```diff
@@ -855,7 +855,7 @@ public function getDB($test = false) {
     }
     else {
       global $CONN;
-      $dsn = 'mysql:dbname=' . $CONN['db'] . ";host=" . $CONN['server'] . ";port=" . $CONN['port'] . ";charset=utf8";
+      $dsn = 'mysql:dbname=' . $CONN['db'] . ";host=" . $CONN['server'] . ";port=" . $CONN['port'] . ";charset=utf8mb4";
```
zyronix, line 858: Add a comment with the following comment:
```
# The utf8mb4 is here to force php to connect with utf8, so you can save emoji's or other non ascii chars into the database. If you are running into issues with this line, we could make this configurable.
```

zyronix (approved):

src/dba/AbstractModelFactory.class.php
zyronix: Also add an item to the changelog.md, you might need to create a header here for the new version.

gochujang-c: Done.

PR #1130, opened 2024-11-21
Bug/999 download file via new api

### LINKED ISSUE #999: [BUG]: Download file via new API
### Version Information

master

### Hashcat

_No response_

### Description

In the old UI it was possible to download a wordlist or a rule file directly from the UI. This is very handy to for example check the contents of those file and maybe create a new version.

This could be done using:
```
http://localhost:8080/getFile.php?file=2
```

As we are moving stuff out of the code base it make sense to add this to the APIv2 aswell. I guess this would be a helper function? But any sugestion on implementation is welcome!

jessevz:
fixes #999

### DIFF
--- a/ci/apiv2/hashtopolis.py
+++ b/ci/apiv2/hashtopolis.py
@@ -818,6 +818,17 @@ class Helper(HashtopolisConnector):
         else:
             return self.resp_to_json(r)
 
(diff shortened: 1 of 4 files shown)


gluafamichl (approved):
Looks okay for me, we will check it locally with the review of the V2 API

Issue #1132, opened 2024-11-21
Adaptation of the API calls for the start page


gpascal123:
![Image](https://github.com/user-attachments/assets/2341a076-6782-42ff-8663-267186f501fb)

Currently, all entries with all fields are always retrieved from the database for agents, tasks, supertasks and cracks.
The frontend then determines the sum of all tasks, for example, and determines which tasks are completed.

The following values are required:

**Agents**
(http://localhost:8080/api/v2/ui/agents?maxResults=3000)
Total of all registered agents and the number of “isActive” agents

**Tasks**
Sum of all tasks and the number of completed tasks
(http://localhost:8080/api/v2/ui/taskwrappers?maxResults=3000&expand=tasks&filter=isArchived=false,taskType=0)
The completed tasks are currently determined in the frontend as follows:

const completedTasks = taskWrapper['tasks'].filter(
  (task) =>
    task.keyspaceProgress === task.keyspace && task.keyspace > 0
);

**Supertasks**
Sum of all supertasks and the number of completed supertasks
(http://localhost:8080/api/v2/ui/taskwrappers?maxResults=3000&expand=tasks&filter=isArchived=false,taskType=1)
The completed supertasks are currently determined in the frontend as follows:

const totalTasks = taskWrapper['tasks'].length;
let completedTasksCount = 0;

taskWrapper['tasks'].forEach((task) => {
  console.log(task);
  if (
    task.keyspaceProgress === task.keyspace &&
    task.keyspace > 0
  ) {
    completedTasksCount++;
  }
});

if (completedTasksCount === totalTasks) {
  completedSTCount++;
}

 **Cracks**

Number of passwords cracked in the last 7 days
(The API call also goes to the hashes and filters “isCracked=true”)
http://localhost:8080/api/v2/ui/hashes?maxResults=3000&filter=isCracked=true
But the result also contains “isCracked=false” entries)

We would like to use either a single API call that only returns the above-mentioned values, or alternatively 4 API calls with the corresponding values each.

jessevz:
So my current idea is to add the count information in the metadata of the getMany API calls. This should be fairly inexpensive since we already get the max value for pagination. The count we can retrieve from the same query. But the downside with this is that the frontend needs to do multiple API requests to get the information.

Now the other solution would be to create a very specific helper function that can get this information. This would be more efficient but this API call would be very tightly coupled with the frontend, and if the frontend changes, so would the API. So I think I like that solution a little bit less.

[...]

PR #1142, opened 2024-11-28
1132 adaptation of the api calls for the start page

jessevz:
solves #1132 I will document what api calls are needed for the home screen when i have some time


gluafamichl:
@jessevz please check again, pytest is failing

PR #1145, opened 2024-12-03
Count

jessevz:
New pull request for count endpoints


jessevz:
This pull request introduces a new count endpoint to enable advanced filtering capabilities, including the ability to filter on expanded (related) objects using inner joins. Additionally, the ORM has been enhanced to support inner joins in count filters, enabling to also filter on the included objects. This was needed so that the correct queries can be made to retrieve the data needed in the start page of the frontend.

gluafamichl (approved):

PR #1174, opened 2025-01-22
Fix: many to many relations from DB did not work

gluafamichl:
This PR fixes the wrong evaluation of many to many relations first seen in superhashlists.

Screenshot from old UI with correct display:
hashlist_old_ui

Screenshot from new UI with dev branch, many to many relations are wrong:
hashlist_new_ui_wrong

Screesnhot from new UI with fixes from this PR, display is correct again:
hashlist_new_ui_correct

See attached JSON files for exact results for this API call:
http://localhost:8080/api/v2/ui/hashlists?maxResults=25&startsAt=0&include=hashType,hashlists&filter[format__eq]=3

Dev-Branch:
[many2one_old.json](https://github.com/user-attachments/files/18508144/many2one_old.json)

Fix from this PR:
[many2many_new.json](https://github.com/user-attachments/files/18508146/many2many_new.json)

### DIFF
--- a/src/inc/apiv2/common/AbstractModelAPI.class.php
+++ b/src/inc/apiv2/common/AbstractModelAPI.class.php
@@ -255,33 +255,32 @@ abstract class AbstractModelAPI extends AbstractBaseAPI
   ): array {
     assert($intermediateFactory instanceof AbstractModelFactory);
     assert($targetFactory instanceof AbstractModelFactory);
-    $retval = array();
-
-
+    $many2Many = array();
+    
     /* Retrieve Parent -> Intermediate -> Target objects */
     $objectIds = [];
-    foreach ($objects as $object) {
+    foreach($objects as $object) {
       $kv = $object->getKeyValueDict();
       $objectIds[] = $kv[$objectField];
     }
     $qF = new ContainFilter($filterField, $objectIds, $intermediateFactory);
     $jF = new JoinFilter($intermediateFactory, $joinField, $joinField);
     $hO = $targetFactory->filter([Factory::FILTER => $qF, Factory::JOIN => $jF]);
-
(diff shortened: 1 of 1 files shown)


src/inc/apiv2/common/AbstractModelAPI.class.php
```diff
@@ -255,33 +255,32 @@ final protected static function getManyToOneRelationViaIntermediate(
...
-    foreach ($hO[$intermediateFactory->getModelName()] as $intermidiateObject) {
-      $kv = $intermidiateObject->getKeyValueDict();
-      $i2p[$kv[$joinField]] = $kv[$filterField];
+    
+    $intermediateObjectList = $hO[$intermediateFactory->getModelName()];
+    $targetObjectList = $hO[$targetFactory->getModelName()];
+    
+    $intermediateObject = current($intermediateObjectList);
+    $targetObject = current($targetObjectList);
+
```
jessevz, line 275: The code will error when an API request gets done for non existing data, because then null will be returned instead of an array. an easy solution would be to initialize an empty array: $many2Many = [];

gluafamichl: You are right, I will fix it.

gluafamichl: What exactly do you mean with non-existing data? I just tried it with no superhashlists defined and it returned an empty array.

jessevz: I made a mistake you are right, the $many2many array gets defined at line 258 so everything is good

gluafamichl: Alright, thank you :-)

PR #1199, opened 2025-03-19
Write cracked hashes to log that have been skipped

jessevz:
Added logging for skipped hashes, so that an administrator can spot the hashes that have been cracked but skipped. This can happen for example when Hashcat rebuilds the hash incorrectly

### DIFF
--- a/src/inc/api/APISendProgress.class.php
+++ b/src/inc/api/APISendProgress.class.php
@@ -231,9 +231,9 @@ class APISendProgress extends APIBasic {
           if (sizeof($hashes) == 0) {
             //This can happen if agent rebuild the hash incorrectly
             //Log the skipped hash so that admin can spot this false negative
-            $logMessage = "Hash has been cracked but skipped! This happened while cracking hashlist with Id: "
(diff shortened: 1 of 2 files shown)


src/inc/api/APISendProgress.class.php
```diff
@@ -229,6 +229,14 @@ public function execute($QUERY = array()) {
           $qF3 = new QueryFilter(Hash::IS_CRACKED, 0, "=");
           $hashes = Factory::getHashFactory()->filter([Factory::FILTER => [$qF1, $qF2, $qF3]]);
           if (sizeof($hashes) == 0) {
+            //This can happen if agent rebuild the hash incorrectly
+            //Log the skipped hash so that admin can spot this false negative
+            $logMessage = "Hash has been cracked but skipped! This happened while cracking hashlist with Id: "
```
s3inlc, line 234: We should make it consistent with the writing of id/Id, I would propose to use the uppercase `ID` writing. Also the `Does` should be written lowercase, and at the end of the log message I would add a dot.

s3inlc (changes requested):
Please add an entry for this in the changelog.md.

s3inlc (approved):

PR #1389, opened 2025-06-24
implemented access group check for hashlists and agents so far

s3inlc:


src/inc/utils/AccessUtils.class.php
```diff
@@ -146,9 +146,9 @@ public static function getAccessGroupsOfUser($user) {
   
   /**
    * @param $agent Agent
-   * @return AccessGroup[]
+   * @return Agent[]
```
jessevz: this should probably be still an array of accesgroups

jessevz (approved):

PR #1458, opened 2025-07-30
Added a notifications class for Microsoft Teams

Drakiat:
Added a notifications class for Microsoft Teams, to be able to send POST notifications to a PowerAutomate webhook to get Hashtopolis logs on a Teams channel or chat. Automation needs to be created in Microsoft Teams under Workflow.


Drakiat:
This is an example workflow as to how it would be set up on Microsoft Teams.
image

ObsidianOracle:
Hello Drakiat,

Thank you so much for your PR. We are currently so involved in Hashtopolis 1.0 last steps that we can´t have a look now. But we will definitely consider it after the 1.0 release. Stay tuned.

Issue #1480, opened 2025-08-06
[BUG]: bug in update script


jessevz:
### Version Information

dev

### Description

Because we changed binarytype to type in agentbinary, old update scripts will crash that expect the old naming. This should be fixed by adding addiitonal checks in these scripts

Issue #1434, opened 2025-07-03
[BUG]: Agents without access group convert to zombies


gluafamichl:
### Version Information

0.14.4 dev

### Description

If you delete an agent from all access groups there is no way in the new UI and API to access this agent. It completely disappears from all agent tables and you can't add it to any access group, it literally rests only as zombie in the database.
In the old version it is still possible to add such agents to an access group again, the same must be added to the new version. The selection field in the edit access groups page should display all agents, which don't belong to the currently edited access group (same behavior like in the old version)

ObsidianOracle:
Only admin should see the agent, if its removed from all access groups

s3inlc:
The issue seems to be with the filter when the frontend is querying the agents to be displayed for selection. It uses the filter_nin (not in) and providing a list of the agents already in the group (which is not wrong at all). 
But for some reason (and this is the bug), the backend then returns an empty list, while there at least the one agent which was removed, should be contained.

jessevz:
> Only admin should see the agent, if its removed from all access groups

@ObsidianOracle so what exactly should we define as admin? The admin user? Or users part of the administrator global permission group? And if so should we not allow admins to reduce the permission of the admin permission group?

PR #1481, opened 2025-08-06
1477 enhancement add hashtypes included by hashcat 7

### LINKED ISSUE #1477: [ENHANCEMENT]: Add hashtypes included by hashcat 7
### Description

With the release of hashcat7 new hashtypes have been added. To support these in hashtopolis they have to be added to the hashtypes database. The new hashtypes can be found in the hashcat 7 release notes section 2: https://github.com/hashcat/hashcat/blob/v7.0.0/docs/releases_notes_v7.0.0.pdf

jessevz:
closes #1477 and #1480


jessevz:
also solves #1434

s3inlc:
The `isSalted` property does not work in the way that it's just determined if there is a salt in a hash or not.

It is only set if there is a salt separated by one colon (default separator) specifically at the end of the hash. 
It is NOT set when a hash has a salt somewhere in all the values present in the hash or if the colon separator is used multiple times (as then it's not the standard separator way anymore (some hashcat modules misuse the separator causing it to split a hash into many pieces) and would break in hashtopolis if we set it to salted as then more than two elements would be created out of a .split(':')

s3inlc:
An example of a hash type which was marked as `isSalted` but should not be: 

 `24000 - BestCrypt v4 Volume Encryption` 
```
$bcve$4$08$323631353538333233323034363039393534383233393530$9f7892b8324b1d8cd36b5f2f8705b407131261620a89370db8369046646f5f82b96780453948db53b04928ae0cc47066f13454b34e31b58ea44ce943bcba14fcbd87f17205a31a896df182629ceea164d87e9e29127e8d865ca0bee52f832723
```

The salt is incorporated into the hash structure directly and it cannot clearly be separated from the hash by a simple split.

Issue #1511, opened 2025-08-14
[ENHANCEMENT]: Backend - Add Apache security headers


gluafamichl:
### Description

Enhancement Description

PR #1513, opened 2025-08-18
1511-enhancement-backend-add-apache-security-headers

gluafamichl:
Add security headers, remove detailed server banner in response


gluafamichl:
Handles issue #1511

PR #1533, opened 2025-08-28
Fix: show archived toggle not updating and persisting after refresh

### LINKED ISSUE #1532: [BUG]: show archived not working
### Version Information

dev

### Hashcat

_No response_

### Description

In the new web ui, when i select show archived, the table doesnt change and also is not responding anymore. also the show archived option is not saved when i refresh the page

RohithMacharla11:
### Bug Fix
- Fixed issue where "Show Archived" toggle did not update the jobs table.
- Added API call to include archived parameter in table fetch.
- Implemented localStorage persistence so toggle state is saved across refreshes.

### Testing
- Verified toggle updates table correctly.
- Verified state persists after page reload.
- Verified no regression in default job table behavior.

Closes #1532


jessevz:
Hi thanks for your pull request, but the issue was refering to the new frontend we are working on (https://github.com/hashtopolis/web-ui). We dont really add new features to the old UI.

PR #1541, opened 2025-09-02
Helper edit tasks

jessevz:
create a helper endpoint that does expensive calculation so that the frontend doesnt have to do these calculations

### DIFF
--- a/src/inc/apiv2/helper/taskExtraDetails.routes.php
+++ b/src/inc/apiv2/helper/taskExtraDetails.routes.php
@@ -70,18 +70,14 @@ class TaskExtraDetailsHelper extends AbstractHelperAPI {
     }
     $task = Factory::getTaskFactory()->get($taskId);
     $estimatedTime = round($timeSpent / ($cProgress / $task->getKeyspace()) - $timeSpent);
-      // $body = $response->getBody();
-      // $body->write($this->ret2json($ret));
+    $currentSpeed = ($currentSpeed > 0) ? Util::nicenum($currentSpeed, 10000, 1000) . "H/s" : 0;
     $responseObject = [
       "estimatedTime" => $estimatedTime,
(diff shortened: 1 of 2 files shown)


src/inc/apiv2/helper/taskExtraDetails.routes.php
```diff
@@ -0,0 +1,106 @@
...
+    if ($taskId == null) {
+      throw new HttpErrorException("No task query param has been provided");
+    }
+    $taskId = intval($taskId);
+    if ($taskId === 0) {
+      throw new HttpErrorException("No valid integer provided as task");
+    }
+
+    $qF = new QueryFilter(Chunk::TASK_ID, $taskId, "=");
+    $chunks = Factory::getChunkFactory()->filter([Factory::FILTER => $qF]);
```
s3inlc, line 53: This will definitely cause performance problems, as when there are larger tasks with many chunks, they are ALL loaded here which will make it slow, put heavy load on the server and probably end in some memory errors.

It would be better to use the sum/max/etc. filters of the DBA to load the required data for this helper.

s3inlc (changes requested):

src/inc/apiv2/helper/taskExtraDetails.routes.php
jessevz: I am not sure if this is really possible with database aggregation. This code is basically copied from /var/www/html/src/tasks.php and the calculations are the same as how the old UI performs them currently

s3inlc: Have a look at the function `getTaskInfo()` and `getChunkInfo()` in `src/inc/Utils.class.php` which are used to load the data for the tasks overview in the old UI, they should pretty much do what you need here I think.

PR #1568, opened 2025-09-05
Remove `version` attribute from docker-compose.yml

injust:
Docker Compose v2.39.2 warns:

```
the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '3.7'
 services:
   hashtopolis-backend:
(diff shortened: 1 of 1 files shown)


s3inlc:
When running on older docker-compose setups, this version attribute would still be required (e.g. when running the newest available docker-compose on Ubuntu 20.04) and there are still users running on older setups.

So I propose to wait some time longer until we remove the version attribute.

PR #1667, opened 2025-10-10
Add: create a new helper to get the speed of a task limited to a max amount to avoid memory overflows and performance problems on edit task page

gluafamichl:
New helper added, which gets the speeds only for the last hour. If more than 10 agents are working on a task, this time period will get smaller.
Please check, if this okay, checked it with the old UI which showed only about 5 minutes.

Frontend changes for this helper are in branch 1663-...


gluafamichl:
As discussed, we can this without any helper in the frontend after detecting a bug there.

Issue #1152, opened 2024-12-09
[ENHANCEMENT]: Implement fetching sparse fieldsets of JSON API standard in new API


jessevz:
### Description

by implementing fetching sparse fieldsets (https://jsonapi.org/format/#fetching-sparse-fieldsets). We can reduce the amount of data that the API needs to send back.

jessevz:
This is especially nice to have for auto refresh, where only specific fields needs to be updated with an interval

PR #1715, opened 2025-10-27
Implemented sparse fieldsets support on the backend #1152

Eric-Wasson:
SF_agent_overview
Agents overview page with sparse fieldsets applied
 
---------------------- 
 
SF_tasks_overview
Tasks overview page with sparse fieldsets applied. A bit more data gets returned than requested. This is because of the aggregatedData function, which doesn't skip the non-included data yet. There was a todo-comment about that already before my commit (now on line 594 in AbstractBaseAPI.class.php). That's something that I'll improve in the future.
 
 ----------------------
 
SF_Tasks_with_several_SF
Task overview page with sparse fieldsets applied for 3 different types of data: task, hashlist and hashType

### DIFF
--- a/src/inc/apiv2/common/AbstractBaseAPI.class.php
+++ b/src/inc/apiv2/common/AbstractBaseAPI.class.php
@@ -602,7 +602,7 @@ abstract class AbstractBaseAPI {
       $attributes[$feature['alias']] = $apiClass::db2json($feature, $kv[$name]);
     }
 
-    $aggregatedData = $apiClass::aggregateData($obj, $aggregateFieldsets);
+    $aggregatedData = $apiClass::aggregateData($obj, $expandResult, $aggregateFieldsets);
     $attributes = array_merge($attributes, $aggregatedData);
     
     /* Build JSON::API relationship resource */
@@ -1197,7 +1197,9 @@ abstract class AbstractBaseAPI {
     array $expands,
     object $object,
     array $expandResult,
-    array $includedResources
+    array $includedResources,
+    array $sparseFieldsets = null,
+    array $aggregateFieldsets = null
 ): array {
     
     // Add missing expands to expands in case they have been added in aggregateData()
@@ -1214,14 +1216,16 @@ abstract class AbstractBaseAPI {
 
       if (is_array($expandResultObject)) {
         foreach ($expandResultObject as $expandObject) {
-          $includedResources = self::addToRelatedResources($includedResources, $apiClass->obj2Resource($expandObject));
+          $noFurtherExpands = [];
+          $includedResources = self::addToRelatedResources($includedResources, $apiClass->obj2Resource($expandObject, $noFurtherExpands, $sparseFieldsets, $aggregateFieldsets));
         }
       } else {
           if ($expandResultObject === null) {
             // to-only relation which is nullable
             continue;
           }
-        $includedResources = self::addToRelatedResources($includedResources, $apiClass->obj2Resource($expandResultObject));
+        $noFurtherExpands = [];
+        $includedResources = self::addToRelatedResources($includedResources, $apiClass->obj2Resource($expandResultObject, $noFurtherExpands, $sparseFieldsets, $aggregateFieldsets));
       }
     }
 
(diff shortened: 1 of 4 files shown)


src/inc/apiv2/common/AbstractBaseAPI.class.php
```diff
@@ -571,6 +569,16 @@ protected function obj2Resource(object $obj, array $expandResult = []): array {
       if ($feature['private'] === true) {
         continue;
       }
+
+      // If sparse fieldsets (https://jsonapi.org/format/#fetching-sparse-fieldsets) is used, return only the requested data
+      if (is_array($sparseFieldsets)) {
```
jessevz: Try to put the three if statements into a single if statement

src/inc/apiv2/model/tasks.routes.php
```diff
@@ -160,36 +160,47 @@ protected function createObject(array $data): int {
...
+    if(is_array($sparseFieldsets) && array_key_exists('task', $sparseFieldsets) && in_array("dispatched", $sparseFieldsets['task']) ) {
+      $aggregatedData["dispatched"] = Util::showperc($keyspaceProgress, $keyspace);
     }
-    
-    //status 1 is running, 2 is idle and 3 is completed
-    $status = 2;
-    if ($keyspaceProgress >= $keyspace && $keyspaceProgress > 0) {
-      $status = 3;
+
+    if(is_array($sparseFieldsets) && array_key_exists('task', $sparseFieldsets) && in_array("searched", $sparseFieldsets['task']) ) {
```
jessevz: the first part of the check:  " if(is_array($sparseFieldsets) && array_key_exists('task', $sparseFieldsets)". is now done twice, I would say do that part as on outer if statement, and then afterwards just check individually for "searched" and "status"

src/inc/apiv2/model/tasks.routes.php
```diff
@@ -160,36 +160,47 @@ protected function createObject(array $data): int {
...
+          $activeAgents[$chunk->getAgentId()] = true;
+        }
+      }
+      
+      $aggregatedData["activeAgents"] = array_keys($activeAgents);
     }
-    elseif (count($activeAgents) > 0) {
-      $status = 1;
+
+    if(is_array($sparseFieldsets) && array_key_exists('task', $sparseFieldsets) && in_array("searched", $sparseFieldsets['task']) ) {
```
jessevz: I think "searched" here, should be "status".

jessevz:
With this code, will it be possible to simply return the entire object and only specify in the sparse fieldset that you also want some extra agregated data? If not I would suggest to use a different query structure for the aggregated data so that you can add aggregated data with just a few extra additions to your query

Eric-Wasson:
Thanks for the feedback! I've fixed the things you mentioned and for the aggregated data it's now possible to query it separately:

Task_overview_with_some_aggregate
Task overview without sparse fieldsets but with selected aggregated data.

--------------------------

Both_SF_and_aggregate
Task overview with sparse fieldsets and selected aggregated data.

--------------------------

SF_without_aggregate
Task overview with sparse fieldsets and without specifying aggregated data fields.

jessevz:
Sorry for the late review, but i see that you use indexes in the examples you provide? like for example fields[agents][0] = agentname. what is the 0 used for? I cant see anything about it in the specs. Also When we want to merge this pull request, we should add a frontend pull request that will query for the aggregations where needed, otherwise some part of the ui will break

Eric-Wasson:
> i see that you use indexes in the examples you provide

Since I used the Firefox devtools I made a slightly wrong implementation of the sparse fieldsets arrays. It's now fixed and can get used properly like this:

`curl --compressed --header "Authorization: Bearer $TOKEN" -g 'http://localhost:8080/api/v2/ui/taskwrappers?page[size]=25&include=accessGroup,tasks,hashlist,hashType&filter[isArchived__eq]=false&fields[task]=attackCmd,isSmall&aggregate[task]=dispatched,status'`

> otherwise some part of the ui will break

What part do you mean would break? As far as I see everything works normally if you don't include the sparse fieldsets and aggregate fieldsets (the default there is to also return everything)  in the query (which the frontend currently doesn't do). If you add the sparse fieldsets or aggregate fieldsets in the request then the response will change too.

jessevz:
Ah I see I thought that the aggregation would only happen if it is defined. That would be my preferred solution. But then I think its fine for now to merge it and later make a pull request to make it so that aggregations only happen when its queried for, and then also do a pull request in the frontend where these queries are altered. Can you fix the open merge conflicts?

jessevz (approved):

Issue #1265, opened 2025-05-01
[ENHANCEMENT]: Properly implement CORS


jessevz:
### Description

Currently we do a CORS hack by just allowing everything, but this should be properly implemented and validated

PR #1725, opened 2025-10-30
Added an improved CORS implementation #1265

Eric-Wasson:
After this has been reviewed and before releasing I would suggest to do two steps:
- Include a hint in the release notes to give a heads up about the new .env variable, so users can add their frontend there after upgrading and don't run into CORS errors.
- Change the default warning in the frontend (_"Network error. Please verify the IP address (localhost:8080) and try again. Note: APIv2 HASHTOPOLIS_APIV2_ENABLE=1 needs to be enabled. "_) to include a hint about the CORS setting (e.g. _"Please make sure to include the used frontend in the .env file."_).

Note: because multiple values in the Allow-Origin header aren't allowed, a small workaround for getting the currently used frontend has been used ([1](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CORS/Errors/CORSMultipleAllowOriginNotAllowed) [2](https://stackoverflow.com/questions/1653308/access-control-allow-origin-multiple-origin-domains))

### DIFF
--- a/doc/installation_guidelines/tls.md
+++ b/doc/installation_guidelines/tls.md
@@ -97,7 +97,7 @@ http {
 }
 ```
 
-3. Update the value of `HASHTOPOLIS_BACKEND_URL` in the `.env` file to reflect the changes done above. 
+3. Update the value of `HASHTOPOLIS_BACKEND_URL` in the `.env` file to reflect the changes done above. If your server name isn't localhost, be sure to also update the comma-separated list of `HASHTOPOLIS_FRONTEND_URLS` to include new https frontend.
 
 4. Start the containers
 ```
--- a/env.example
+++ b/env.example
@@ -9,4 +9,4 @@ HASHTOPOLIS_DB_HOST=db
 
 HASHTOPOLIS_APIV2_ENABLE=0
 HASHTOPOLIS_BACKEND_URL=http://localhost:8080/api/v2
-HASHTOPOLIS_FRONTEND_URLS=http://127.0.0.1:4200,https://127.0.0.1:4200,http://localhost:4200,https://localhost:4200,http://127.0.0.1:8080,https://127.0.0.1:8080,http://localhost:8080,https://localhost:8080
+HASHTOPOLIS_FRONTEND_URLS=http://127.0.0.1:4200,http://localhost:4200,http://127.0.0.1:8080,http://localhost:8080,https://127.0.0.1,https://localhost


src/api/v2/index.php
```diff
@@ -190,7 +188,19 @@ public static function addCORSheaders(Request $request, $response) {
     $methods = $routingResults->getAllowedMethods();
     $requestHeaders = $request->getHeaderLine('Access-Control-Request-Headers');
     
-    $response = $response->withHeader('Access-Control-Allow-Origin', '*');
+    if (getenv('HASHTOPOLIS_FRONTEND_URLS') !== false) {
+      if(in_array($_SERVER['HTTP_ORIGIN'], explode(',', getenv('HASHTOPOLIS_FRONTEND_URLS')), true)) {
```
jessevz: In the SLIM framework you dont access the ehaders with $_SERVER. replace this with $request->getHeaderLine('HTTP_ORIGIN')

src/api/v2/index.php
```diff
@@ -190,7 +188,19 @@ public static function addCORSheaders(Request $request, $response) {
...
+        $response = $response->withHeader('Access-Control-Allow-Origin', $_SERVER['HTTP_ORIGIN']);
```
jessevz: The same here, user getHeaderLine()

src/api/v2/index.php
```diff
@@ -190,7 +188,19 @@ public static function addCORSheaders(Request $request, $response) {
...
+      }
+      else {
+        Util::createLogEntry(DLogEntryIssuer::USER, Login::getInstance()->getUserID(), DLogEntry::WARN, "CORS error: Allow-Origin doesn't match. Please make sure to include the used frontend in the .env file.");
```
jessevz: I dont think this type of logging makes sense here, especially since an user might not be logged in yet before he runs into cors errors. I think using error_log() makes more sense

src/api/v2/index.php
```diff
@@ -190,7 +188,19 @@ public static function addCORSheaders(Request $request, $response) {
...
+      }
+    }
+    else {
+      //No frontend URLs given in .env file, switch to default allow all
+      $response = $response->withHeader('Access-Control-Allow-Origin', '*');
+    }
+
     $response = $response->withHeader('Access-Control-Allow-Methods', implode(',', $methods));
```
jessevz, line 204: Probably better to allow only the methods that hashtopolis needs, which I think are: HEAD, OPTIONS, GET, PUT, PATCH, POST

Eric-Wasson: That's seems to be already implemented, right now only the allowed headers for each resource get returned, e.g. for the task with id 1:
Task_allowed_headers

Other example:
Other_example

PR #1768, opened 2025-11-20
Added backend handling of overwriting on import pre-cracked hashes and import by file and URL

Eric-Wasson:
Added backend handling of overwriting the plaintext on importing pre-cracked hashes as well as upload pre-cracked hashlists and upload using URL.

UI todos before merge:
- A new checkbox input field on the new UI is needed (checkbox name: "overwrite"). Because the checkbox is currently missing, a "Required parameter not specified" error is thrown when trying to import pre-cracked hashes.
- The separator is hardcoded to ":", the user should be able to set it with an input field like on the old UI (text input field with the name "separator").
- Use the same handling that's used for the create new hashlist, i.e. the upload file option does a POST-request to importFile first to create the file, then uses PATCH to add the data and then does the POST to importCrackedHashes to import the uploaded file. Also the upload by URL can be added on the frontend now (but currently that's not implemented on the create hashlist frontend either).

### DIFF
--- a/src/inc/apiv2/helper/importCrackedHashes.routes.php
+++ b/src/inc/apiv2/helper/importCrackedHashes.routes.php
@@ -3,6 +3,8 @@
 use DBA\Hash;
 use DBA\Hashlist;
 
+use Middlewares\Utils\HttpErrorException;
+
 require_once(dirname(__FILE__) . "/../common/AbstractHelperAPI.class.php");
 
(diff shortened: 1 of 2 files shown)


jessevz:
This pull request crashes in the build process because of an update to php 8.5. in the dev branch i have pinned it to 8.4. if you merge the dev branch into this branch, it should pass the build process

Issue #1799, opened 2025-12-01
[BUG]: Legacy Agent Type checking not working properly


s3inlc:
### Version Information

rainbow4

### Description

The legacy binary type check used in the older update scripts does not work due to the reason that when the DBA function set() is used, it does a get() at the end, where it again will use the new structure of the model causing a stack trace:

```
PHP Fatal error:  Uncaught PDOException: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'binaryType' in 'field list' in /var/www/html/src/dba/AbstractModelFactory.class.php:538
Stack trace:
#0 /var/www/html/src/dba/AbstractModelFactory.class.php(538): PDOStatement->execute()
#1 /var/www/html/src/dba/AbstractModelFactory.class.php(518): DBA\AbstractModelFactory->getFromDB()
#2 /var/www/html/src/dba/models/AgentBinaryFactory.class.php(79): DBA\AbstractModelFactory->get()
#3 /var/www/html/src/dba/AbstractModelFactory.class.php(309): DBA\AgentBinaryFactory->get()
#4 /var/www/html/src/inc/Util.class.php(219): DBA\AbstractModelFactory->set()
#5 /var/www/html/src/install/updates/update_v0.14.3_v0.14.4.php(8): Util::checkAgentVersionLegacy()
#6 /var/www/html/src/install/updates/update.php(54): include('...')
#7 /var/www/html/src/inc/load.php(82): include('...')
#8 {main}
  thrown in /var/www/html/src/dba/AbstractModelFactory.class.php on line 538
```

PR #1802, opened 2025-12-02
Fixed bug in legacy agentbinary update, by directly performing update…

jessevz:
… logic in order to skip get() which result in bug


s3inlc:
closes #1799

PR #1807, opened 2025-12-03
Hashtopolis lib extracted

s3inlc:
As we want to be able to use the functionality which was in hashtopolis.py also standalone separately and also allowing user to use it in their own python scripts and setups, it is outsourced to a separate repository and included via pip.

### DIFF
--- a/ci/apiv2/requirements.txt
+++ b/ci/apiv2/requirements.txt
@@ -2,3 +2,4 @@ click
 click_log
 pytest
 git+https://github.com/hashtopolis/python-hashtopolis.git
+confidence


ci/apiv2/requirements.txt
```diff
@@ -1,5 +1,4 @@
 click
 click_log
-confidence==0.17
```
jessevz, line 3: I think conifdence is still a requirement. for example in ci/apiv2/hastopolis_agent.py

jessevz: But I see that the pytestst are still performed so maybe it is not needed

s3inlc: ah good point, then I should leave it in there. Tests were still working as it is a dependency of the hashtopolis lib, and therefore gets installed as well.

PR #1811, opened 2025-12-05
Added missing CORS variables to two config files

Eric-Wasson:
Added missing CORS variables to two config files


s3inlc (approved):
I think it's not ideal that we have to put in that many default values, but for now I would merge it and in a refactoring step we have to explore how we can derive this value automatically from the backend url.

PR #1826, opened 2025-12-10
Made dockerfile smaller by using smaller slim base image

jessevz:
Since we only use the prebuilt image for compiling sqlx-cli, we can get away with a smaller image


s3inlc:
Technically this size change only is relevant when the image is built, as it needs to pull less data, but it does not influence the image size of the actual hashtopolis image. But why not, it's at least something :)

PR #1833, opened 2025-12-11
Made it possible to sort on 1 to 1 relationship attributes

### LINKED ISSUE #1832: [ENHANCEMENT]: Sort on included objects api
### Description

Make it possible to sort on included data in the API so that its possible to click on every column for sorting in the frontend.

jessevz:
closes #1832

### DIFF
--- a/src/dba/CoalesceOrderFilter.class.php
+++ b/src/dba/CoalesceOrderFilter.class.php
@@ -13,6 +13,10 @@ class CoalesceOrderFilter extends Order {
   }
   
   function getQueryString(AbstractModelFactory $factory, bool $includeTable = false): string {
-    return "COALESCE(" . implode(", ", $this->columns) . ") " . $this->type;
+    $mapped_columns = [];
+    foreach($this->columns as $column) {
+      array_push($mapped_columns, AbstractModelFactory::getMappedModelKey($factory->getNullObject(), $column));
+    }
+    return "COALESCE(" . implode(", ", $mapped_columns) . ") " . $this->type;
   }
 }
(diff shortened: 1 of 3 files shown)


src/dba/CoalesceOrderFilter.class.php
```diff
@@ -0,0 +1,18 @@
...
+  private $columns;
+  private $type;
+  
+  function __construct($columns, $type) {
+    $this->columns = $columns;
+    $this->type = $type;
+  }
+  
+  function getQueryString(AbstractModelFactory $factory, bool $includeTable = false): string {
+    return "COALESCE(" . implode(", ", $this->columns) . ") " . $this->type;
```
s3inlc: To ensure that there will not be any problem on mapped columns, we have to ensure that the column names in $this->columns are translated similar for example how it's done in the standard order filter. For each of the column this mapping translation has to be checked similar as this:
```
AbstractModelFactory::getMappedModelKey($factory->getNullObject(), $column);
```

src/dba/JoinFilter.class.php
```diff
@@ -27,18 +27,31 @@ class JoinFilter extends Join {
...
   /**
    * JoinFilter constructor.
    * @param $otherFactory AbstractModelFactory
    * @param $matching1 string
    * @param $matching2 string
    * @param $overrideOwnFactory AbstractModelFactory
+   * @param $joinType string is normally inner, left or right
    */
-  function __construct($otherFactory, $matching1, $matching2, $overrideOwnFactory = null) {
+  function __construct($otherFactory, $matching1, $matching2, $overrideOwnFactory = null, $joinType = "inner", $queryFilters = []) {
```
s3inlc: in order to keep consistent on how we build the queries I would write the SQL instruction all uppercase (i.e. `INNER`). Also I think it would make sense to add constant strings to the JoinFilter class so that we don't have to put in strings in the argument but for example: `JoinFilter::INNER` instead.

PR #1841, opened 2025-12-17
Read DB port env variable properly in db check in entrypoint

### LINKED ISSUE #1840: [BUG]: docker entrypoint only works with database default ports
### Version Information

dev

### Hashcat

_No response_

### Description

The check if the database is ready in docker-entrypoint.sh is only working properly if the default port for the corresponding database is used. In the mysql/psql commands the ports are not provided.

s3inlc:
closes #1840

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -18,10 +18,16 @@ if [[ "$HASHTOPOLIS_DB_TYPE" == "mysql" ]]; then
     echo "Using MySQL..."
     DB_CMD="mysql -u${HASHTOPOLIS_DB_USER} -p${HASHTOPOLIS_DB_PASS} -h ${HASHTOPOLIS_DB_HOST} --skip-ssl"
     DB_TYPE="mysql"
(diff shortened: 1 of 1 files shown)


docker-entrypoint.sh
```diff
@@ -18,10 +18,16 @@ if [[ "$HASHTOPOLIS_DB_TYPE" == "mysql" ]]; then
     echo "Using MySQL..."
     DB_CMD="mysql -u${HASHTOPOLIS_DB_USER} -p${HASHTOPOLIS_DB_PASS} -h ${HASHTOPOLIS_DB_HOST} --skip-ssl"
     DB_TYPE="mysql"
+    if [[ -n "${HASHTOPOLIS_DB_PORT}" ]]; then
```
jessevz, line 21: This variable should probably also be added to the env.example

s3inlc: I did not include it in the env example as there it takes the default ports without configuration and therefore is only optional (and probably very rarely used).

jessevz (approved):

PR #1851, opened 2025-12-28
Optimized check for already cracked hashes

Grr1337:
The check, whether hashes from a new hashlist are already cracked in another list, got optimized.
Instead of searching each hashline in the database, the entire comparion got sourced out to the database.
A temporary table of the new hashes is created and via join cracked hashes will be determined.

This will speed up the comparison for lists of a bigger size.

I haven't done comprehensive performance tests, but with my hashtopolis hash table containing 10 million rows and a 500k lines hashfile, I achieved 40 seconds instead of 1 minute and 15 seconds.

### DIFF
--- a/src/inc/utils/HashlistUtils.class.php
+++ b/src/inc/utils/HashlistUtils.class.php
@@ -852,7 +852,7 @@ class HashlistUtils {
         $db = Factory::getAgentFactory()->getDB();
         $db->query(
           "CREATE TEMPORARY TABLE tmp_hashes (
-            hash VARCHAR(255)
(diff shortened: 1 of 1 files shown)


s3inlc:
Thanks for your pull request. Please note that currently the big changes are happening in the dev branch and besides some bugfixes, the master will be updated at the point where the dev gets merged.
The issue with your optimization here is, that you are using in-line MySQL which makes it inflexible for other systems (we have now MySQL and PostgreSQL supported in the current development). 
If it's possible to make this temporary table database system agnostic, it could be interesting though.

PR #1877, opened 2026-01-22
Added helper to retrieve files in the import directory

jessevz:


src/inc/apiv2/helper/importFile.routes.php
```diff
@@ -386,6 +386,31 @@ function processDelete(Request $request, Response $response, array $args): Respo
...
+   * Scans the import-directory for files. Directories are ignored.
+   * @return array of all files in the top-level directory /../import
+   */
+  function scanImportDirectory() {
+    $directory = Factory::getStoredValueFactory()->get(DDirectories::IMPORT)->getVal() . "/";
+    if (file_exists($directory) && is_dir($directory)) {
+      $importDirectory = opendir($directory);
+      $importFiles = array();
+      while ($file = readdir($importDirectory)) {
+        if ($file[0] != '.' && $file != "." && $file != ".." && !is_dir($file)) {
```
s3inlc: I think the part `&& $file != "." && $file != ".."` is redundant, as the first character is check for a dot which both of the other two cases have as well.

PR #1884, opened 2026-01-22
Test pipeline

jessevz:


jessevz:
Static code analyzer added to the github actions, for now will scan only the new api code, but once a lot of legacy code has been removed we can scan more directories and also increase the level of the scanning

PR #1917, opened 2026-02-05
Adding additional indexes

s3inlc:
This adds all missing indexes on foreign keys in postgres (as in mysql they are created by default, but not in postgres).

Additionally, we create indexes on (isArchived, priority) for the tables Task and TaskWrapper, and on isArchived for the table Hashlist. Unfortunately, it is not possible to use `IF NOT EXISTS` on mysql create and drop index statements, therefore I had to create the stored procedures to avoid people having issues with the migrations if some index already exists or it tries to drop an index which does not exist.


src/migrations/mysql/20260205140244_indexes.sql
```diff
@@ -0,0 +1,60 @@
...
+        PREPARE st FROM @sqlstmt;
+        EXECUTE st;
+        DEALLOCATE PREPARE st;
+    END IF;
+
+END;
+
+-- create new indexes on some isArchived columns which is used on a lot of queries
+CALL CreateIndex('Hashlist', 'isArchived', 'isArchived, hashlistId');
+CALL CreateIndex('Task', 'isArchived_priority', 'isArchived, priority');
```
jessevz: I think as a third index, the composite index should also have the taskId

src/migrations/mysql/20260205140244_indexes.sql
```diff
@@ -0,0 +1,60 @@
...
+
+CALL DropIndex('TaskWrapper', 'isArchived'); -- we drop and replace the single isArchived index with the following composite one
+CALL CreateIndex('TaskWrapper', 'isArchived_priority', 'isArchived, priority');
```
jessevz: The same here, the TaskWrapperId as third composite index

jessevz:
Some additional index that we might want to add:

I created the following index a while back:

  ADD KEY `isCracked` (`isCracked`),

But I think it works better for us if its:

  ADD KEY `isCracked` (`isCracked`, `timeCracked` DESC, `HashId` DESC),

src/migrations/mysql/20260205140244_indexes.sql
s3inlc: is this to cover the specific `SELECT * FROM Task WHERE isArchived=0 ORDER BY priority DESC, taskId ASC` statements?

jessevz: yes, the taskId is a tie breaker. And especially since priority it is easy to have priority with the same value, we also need the taskId in the index. That being said, currently I think taskid and priority are either both ASc or both DESC. we still have to change this and then also reflect this then in the index

s3inlc: I think the index does not really care about the ordering, e.g. when you do DESC it will just start from the other direction on the index

jessevz: Yes when both indexes are the same (both ASC for example) this is possible, but when you have one index that is ASC and the other is DESC this is not possible, because you cant just start from the other direction anymore

s3inlc: but that would mean with the double composite (isArchived, priority) you could do queries on both ascending and descending priority, but with the three key index (isArchived, priority, taskId) you could not use it for the query: `SELECT * FROM Task WHERE isArchived=0 ORDER BY priority ASC, taskId ASC `, correct?

From what I read it's not recommended that much to extensively use composite indexes, especially with three keys, probably due to the fact that you loose some coverage on other queries where the index then cannot be used anymore.

s3inlc: so it seems in our case it makes sense to do three columns, I will add them for both Task and TaskWrapper..

jessevz (approved):

PR #1925, opened 2026-02-12
Credentials/Startup Refactoring for testing and DBA Column Filter

s3inlc:
Removed all global variables like `$CONN`, `DBA_xx`, `$VERSION` into a StartupConfig singleton class to avoid issues in order of includes. 
The only remaining ones are legacy related and supposed to be removed in the future.

This also allowed to get rid of `confv2.php` and `info.php` which are now incorporated into `StartupConfig`.

### DIFF
--- a/src/dba/AbstractModelFactory.class.php
+++ b/src/dba/AbstractModelFactory.class.php
@@ -487,12 +487,7 @@ abstract class AbstractModelFactory {
     $stmt = $dbh->prepare($query);
     $stmt->execute($vals);
     
-    // put result into one array
-    $result = [];
-    while ($row = $stmt->fetch(PDO::FETCH_NUM)) {
-      $result[] = $row[0];
-    }
-    return $result;
+    return $stmt->fetchAll(PDO::FETCH_COLUMN);
   }
   
   public function sumFilter($options, $sumColumn) {


src/dba/AbstractModelFactory.class.php
```diff
@@ -463,6 +465,36 @@ public function multicolAggregationFilter($options, $aggregations) {
...
+   * @param $column string single column key which should be retrieved
+   * @return array of the column entries returned from this query
+   */
+  public function columnFilter(array $options, string $column): array {
+    $query = "SELECT " . Util::createPrefixedString($this->getMappedModelTable(), [self::getMappedModelKey($this->getNullObject(), $column)]);
+    $query = $query . " FROM " . $this->getMappedModelTable();
+    
+    $vals = array();
+    
+    if (array_key_exists(Factory::JOIN, $options)) {
```
jessevz, line 479: Do joins make sense when you need a single column? I would assume you just use the single factory you need

src/dba/AbstractModelFactory.class.php
```diff
@@ -463,6 +465,36 @@ public function multicolAggregationFilter($options, $aggregations) {
...
+      $query .= $this->applyFilters($vals, $options[Factory::FILTER]);
+    }
+    
+    $dbh = self::getDB();
+    $stmt = $dbh->prepare($query);
+    $stmt->execute($vals);
+    
+    // put result into one array
+    $result = [];
+    while ($row = $stmt->fetch(PDO::FETCH_NUM)) {
```
jessevz: instead of doing this with a loop, I think you can do PDO::FETCH_COLUMN, and that will already return the results in the format you want

src/dba/AbstractModelFactory.class.php
s3inlc: Yes, for example when you want to retrieve all hashlists IDs you have access to. You would need a join with accessgroup, but in the end you are just interested in the hashlist IDs.

src/dba/AbstractModelFactory.class.php
s3inlc: I will check

PR #1946, opened 2026-02-20
Update the basic install manual according to the latest release

coiseiw:
Adjusting the manual for the installation since the security update


ObsidianOracle (approved):
Fine.

PR #1953, opened 2026-03-04
Added helper for getting available tasks for agent

jessevz:


src/inc/apiv2/helper/GetBestTasksAgent.php
```diff
@@ -0,0 +1,99 @@
...
+  }
+  
+  static public function register($app): void {
+    $baseUri = GetBestTasksAgent::getBaseUri();
+    
+    /* Allow CORS preflight requests */
+    $app->options($baseUri, function (Request $request, Response $response): Response {
+      return $response;
+    });
+    $app->get($baseUri, "Hashtopolis\\inc\\apiv2\\helper\\getBestTasksAgent:handleGet");
```
s3inlc: I think here the class name should be used which would need to be capitalized, maybe this got forgotten on the copilot suggestion change?

s3inlc (changes requested):

s3inlc (approved):

PR #1965, opened 2026-03-11
Api tokens

jessevz:


src/migrations/mysql/20260309164000_api-key.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE JwtApiKey (
+    JwtApiKeyId INT NOT NULL AUTO_INCREMENT,
+    userId INTEGER,
```
s3inlc, line 3: additional to the suggestions of copilot, userId and isRevoked probably also should have 'NOT NULL'

src/migrations/postgres/20260309164000_api-key.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE JwtApiKey (
+    JwtApiKeyId SERIAL NOT NULL PRIMARY KEY,
+    userId INTEGER,
```
s3inlc, line 3: additional to the suggestions of copilot, userId and isRevoked probably also should have 'NOT NULL'

src/migrations/mysql/20260309164000_api-key.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE JwtApiKey (
+    JwtApiKeyId INT NOT NULL AUTO_INCREMENT,
```
s3inlc: please use camelcase with a lowercase start as this is done on all column names for the db

src/migrations/postgres/20260309164000_api-key.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE JwtApiKey (
+    JwtApiKeyId SERIAL NOT NULL PRIMARY KEY,
```
s3inlc: please use camelcase with a lowercase start as this is done on all column names for the db

src/dba/models/generator.php
```diff
@@ -287,6 +287,15 @@
     ['name' => 'errors', 'read_only' => True, 'type' => 'str(65535)', 'protected' => True],
   ],
 ];
+$CONF['JwtApiKey'] = [
+  'columns' => [
+    ['name' => 'jwtApiKeyId', 'read_only' => True, 'type' => 'int', 'protected' => True],
```
s3inlc, line 292: the naming of the column key is correct here, but was it maybe not re-generated as it is capitalized in the model classes?

s3inlc (changes requested):

src/migrations/mysql/20260309164000_api-key.sql
jessevz: Agree with isRevoked, but userId should be nullable, because otherwise you cant delete an user when it still has an api token because of foreignkey constraints. I will add that when an user is deleted, it will turn the userid of the apikey to null

s3inlc: Okay, if this is the process which should be done.
Is it ensured then that such tokens which are not assigned anymore can be handled by the admin? Otherwise we would then up with tokens floating around which no one can see.

src/migrations/mysql/20260309164000_api-key.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE JwtApiKey (
+    jwtApiKeyId INT NOT NULL AUTO_INCREMENT,
+    userId INTEGER,
+    startValid bigint NOT NULL,
+    endValid bigint NOT NULL,
+    isRevoked BOOLEAN NOT NULL DEFAULT FALSE,
+    PRIMARY KEY (`jwtApiKeyId`),
+    KEY `idx_JwtApiKey_userId` (`userId`),
```
s3inlc: as we renamed the column to be non-capitalized, I would keep it consistent also in the index/key names

src/migrations/postgres/20260309164000_api-key.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE JwtApiKey (
+    jwtApiKeyId SERIAL NOT NULL PRIMARY KEY,
+    userId INTEGER,
+    startValid bigint NOT NULL,
+    endValid bigint NOT NULL,
+    isRevoked BOOLEAN NOT NULL DEFAULT FALSE,
+    CONSTRAINT fk_JwtApiKey_user
```
s3inlc, line 7: as we renamed the column to be non-capitalized, I would keep it consistent also in the index/key names

src/migrations/postgres/20260309164000_api-key.sql
```diff
@@ -0,0 +1,10 @@
...
+        FOREIGN KEY (userId) REFERENCES htp_User(userId)
+);
+CREATE INDEX idx_JwtApiKey_userId ON JwtApiKey (userId);
```
s3inlc: as we renamed the column to be non-capitalized, I would keep it consistent also in the index/key names

s3inlc (changes requested):

src/migrations/mysql/20260309164000_api-key.sql
jessevz: Yes, admin has access to all tokens

s3inlc (approved):

PR #1992, opened 2026-03-17
Removed rule splitting

### LINKED ISSUE #705: Remove Rule Splitting
As with the introduction of the -S flag in hashcat the rule splitting feature in Hashtopolis became I obsolete, I suggest removing it. It causes issues with normal tasks more than it should.

jessevz:
closes #705


src/inc/defines/DConfig.php
s3inlc: I think these should be removed as well, no?

s3inlc (changes requested):

src/inc/defines/DConfig.php
jessevz: Yes you are right, i will remove them

s3inlc (approved):

PR #2024, opened 2026-03-26
Fix api docs

jessevz:


jessevz:
@s3inlc @correct-horse-battery-bench Only reason the the checks failed previously was that a repo that was needed was temporarily unavailable. This can be reviewed now

s3inlc (approved):

correct-horse-battery-bench:
@s3inlc @jessevz Added redocly which checks if openapi compliant, spectral just checks if json+api spec compliant. As there are currently still openapi errors modified the pipeline to not fail on error. Can be merged but openapi errors would need remedy at some point.

PR #2033, opened 2026-04-10
Also check if the total hash count of a hashlist needs to be fixed

s3inlc:
So far only the cracked hashes counts where check, but not the total hash counts of the hashlist.


jessevz (approved):

jessevz:
We could reduce the amount of database queries here by instead of doing Factory::getHashlistFactory()->set(....) when a count doesnt match, is by adding the field that needs to be updated to an array, and then afterwards do mset if the array is not empty, then both hashcount and cracked can be updated in a single query. But this is a very minor performance improvement so I approved this pr

s3inlc:
This action is one of the candidates we will have to consider for background tasks anyway when we are going to add something like this. I would consider in such a refactoring then to optimize the implementation.

PR #2053, opened 2026-05-05
2044 bug missing color labeling of tasks

inadequate777:
* added missing fields for color in taskwrapperdisplay


inadequate777:
need to add some test before merge

inadequate777:
Added test for color field

ci/apiv2/test_taskwrapperdisplays.py
```diff
@@ -0,0 +1,35 @@
...
+
+class TaskWrapperDisplaysTest(BaseTest):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.config = HashtopolisConfig()        
+    
+    def test_taskwrapperdisplays_returns_color_field(self):
+        hashlist = self.create_hashlist()
+        task = self.create_task(hashlist, extra_payload={'color': '#8000ff'})
```
jessevz: the create_task files already contain a color (ex. /var/www/html/ci/apiv2/testfiles/task/create_task_001.json). Use that instead of the extra_payload

ci/apiv2/test_taskwrapperdisplay.py
```diff
@@ -0,0 +1,35 @@
...
+        cls.config = HashtopolisConfig()        
+    
+    def test_taskwrapperdisplays_returns_color_field(self):
+        hashlist = self.create_hashlist()
+        task = self.create_task(hashlist, extra_payload={'color': '#8000ff'})
+        conn = HashtopolisConnector('/ui/taskwrapperdisplays', self.config)
+        conn.authenticate()
+
+        headers = conn._headers
+        uri = conn._api_endpoint + conn._model_uri
```
jessevz: Instead of these steps that setup a manual connection, an object should be taskwrapperdisplay object should be made in https://github.com/hashtopolis/python-hashtopolis to be reused and more consistent with the other tests

jessevz (changes requested):

inadequate777:
* Depends on https://github.com/hashtopolis/python-hashtopolis/pull/3

inadequate777:
* tests work after python-hashtopolis merge

src/migrations/mysql/20260413140000_task-view.sql
```diff
@@ -2,7 +2,7 @@ CREATE VIEW TaskWrapperDisplay AS SELECT
     tw.taskWrapperId AS taskWrapperId, tw.priority AS taskWrapperPriority, tw.maxAgents AS taskWrapperMaxAgents,
     tw.taskType AS taskType, tw.hashlistId AS hashlistId, tw.accessGroupId AS accessGroupId,
     tw.taskWrapperName AS taskWrapperName, tw.isArchived AS taskWrapperIsArchived, tw.cracked AS cracked,
-    t.taskId AS taskId, t.taskName AS taskName, t.attackCmd AS attackCmd, t.chunkTime AS chunkTime,
```
jessevz, line 5: You cant change a migration that has already happened, this way systems that have performed the migration will end up in a bad state. You should create a new migration where you add this field

jessevz (changes requested):

inadequate777:
added new files for updating taskwrapperdisplay view using CREATE OR REPLACE VIEW in both sql and postres

jessevz (approved):

PR #2054, opened 2026-05-05
1161 bug new api duplicate links object in get one api responses

osaerdna:
- Moved the data self reference from `AbstractBaseAPI::obj2Resource` to `AbstractModelAPI::getManyResources`
- Added tests for getOne, getMany and POST requests in `test_http_methods.py`

I have made some manual testing in the web-ui and cant see that anything breaks with the change.

@jessevz, please add other reviewer as you see fit.

### DIFF
--- a/ci/apiv2/test_http_methods.py
+++ b/ci/apiv2/test_http_methods.py
@@ -1,7 +1,10 @@
 import requests
+import json
+import time
 
 from hashtopolis import HashtopolisConnector, HashtopolisConfig
 from utils import BaseTest
+
 
 
 class HttpMethodsTest(BaseTest):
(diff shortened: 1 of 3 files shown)


osaerdna:
The change in the api (no self reference inside the data entity) break the implementation of python-hashtopolis that relies on that one when creating objects. I will check if there is a way to implement a backward compatible fix, otherwise I would appreciate some guidance on how to move forward with this one.

jessevz:
I have looked at the code and I think it might be better after all to keep the duplicate uri. it is only for getOne requests and by keeping it it is easier for clients to use the api and not have to switch between to many to one cases.  Just like we see now in the hashtopolis.py case

osaerdna:
I agree with that. I will close this pr and update the ticket with this info. Thanks.

osaerdna:
Changed requirements, removing the links data property break existing clients and is not worth it at the moment.

PR #2057, opened 2026-05-07
1973 bug edit superhashlist bug in hashlist datasource

prettyaveragecoder:
## Summary
Fix missing meta.page.total_elements in getOneResource for to-many includes.

Frontend got a TypeError when editing a superhashlist because
response.meta.page.total_elements was undefined in the API response.

## Changes
- AbstractBaseAPI.php: getOneResource now returns meta.page.total_elements
  when the request includes a to-many relationship (e.g. include=hashlists)

## Known limitation
The current implementation counts total expanded objects across all to-many
relationships in a single value. This works correctly today because no
endpoint includes multiple to-many relationships simultaneously. If that
changes in the future, total_elements may return an incorrect combined count
and will need to be revisited.

## Test
Added test_superhashlist_edit_returns_meta_page_total_elements in
test_hashlist.py to verify that meta.page.total_elements is returned
correctly when fetching a superhashlist with include=hashlists.

The test makes a direct HTTP request instead of using the hashtopolis
Python library, because the library only returns the deserialized data
object (response['data']) and discards the meta field. By using
requests.get() directly we get the raw JSON:API response and can assert
that meta.page.total_elements exists and has the correct value.

### DIFF
--- a/ci/apiv2/test_hashlist.py
+++ b/ci/apiv2/test_hashlist.py
@@ -1,4 +1,6 @@
-from hashtopolis import Hashlist, Helper, File
+import requests
+
+from hashtopolis import Hashlist, Helper, File, HashtopolisConnector, HashtopolisConfig
 from utils import BaseTest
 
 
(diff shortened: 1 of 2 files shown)


jessevz:
How does this work when you are included multiple to many expands? Wont the last always overwrite the result?

Personally I would say to solve this is to not load a pagesize for this in the frontend since pagination is not possible for the included section anyway

PR #2067, opened 2026-05-12
Get correct intersection of legacy api permissions instead of new CRU…

correct-horse-battery-bench:
…D permissions with legacy permissions

Currently if admin creates a jwt key and specifies the scopes in the request the scopes are ignored and scope "[]" is set.

Previously `array_intersect_key($rightgroup_perms, array_flip($scopes))` compared legacy permissions with new CRUD scopes - >always empty.

Tests were generated.

### DIFF
--- a/src/inc/apiv2/model/ApiTokenAPI.php
+++ b/src/inc/apiv2/model/ApiTokenAPI.php
@@ -79,7 +79,7 @@ class ApiTokenAPI extends AbstractModelAPI {
       $this->getRightGroup($this->getCurrentUser()->getRightGroupId())->getPermissions()
     );
 
-    // Modern CRUD scope dict: true iff the perm was requested AND the user has it.
+    // Modern CRUD scope dict: true if the perm was requested AND the user has it.
     $requestedScopes = [];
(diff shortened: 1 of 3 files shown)


s3inlc (approved):

src/inc/apiv2/model/ApiTokenAPI.php
```diff
@@ -74,19 +75,15 @@ protected function createObject(array $data): int {
...
-      $rightgroup_perms = json_decode($allPermissions, true);
-    }
-    $NotAllowedPerms = array_filter($rightgroup_perms, fn($v) => $v === false);
-    $allowedPerms = array_intersect_key($rightgroup_perms, array_flip($scopes));
+    $userCrudPerms = AccessUtils::getPermissionArrayConverted(
+      $this->getRightGroup($this->getCurrentUser()->getRightGroupId())->getPermissions()
+    );
 
-    $requestedScopes = $allowedPerms + $NotAllowedPerms;
+    // Modern CRUD scope dict: true iff the perm was requested AND the user has it.
```
jessevz: typo: iff should be if

jessevz (approved):

PR #2093, opened 2026-05-19
fix: resolve task sorting error by mapping displayName alias to taskWrapperName

alricium:
The task overview page was failing to sort by name because the frontend sends `displayName` as the sorting parameter, but the backend expected `taskWrapperName`.

This PR updates the `TaskWrapper` model to map the `displayName` alias to the `taskWrapperName` column. This ensures that the backend's dynamic filtering engine can correctly translate the sorting request into the appropriate database query.


s3inlc:
Everything currently developed for the new frontend is done in the dev branch, the master is only for the current UI and bugfix releases. Also it does not make sense to just rename the alias, because the attributes to be changed should be in the View used for the table, not the TaskWrapper table.

PR #2099, opened 2026-05-20
2074 bug chunk count missing on task details and percentage sign missing

inadequate777:
* Returns total number of chunks as aggregate

### DIFF
--- a/src/inc/apiv2/model/TaskAPI.php
+++ b/src/inc/apiv2/model/TaskAPI.php
@@ -214,7 +214,11 @@ class TaskAPI extends AbstractModelAPI {
         $chunks = Factory::getChunkFactory()->filter([Factory::FILTER => $qF]);
         $aggregatedData["status"] = TaskUtils::getStatus($chunks, $keyspace, $keyspaceProgress);
       }
-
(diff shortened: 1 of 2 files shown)


s3inlc (approved):

src/inc/apiv2/model/TaskAPI.php
```diff
@@ -214,7 +214,10 @@ function aggregateData(object $object, array &$included_data = [], ?array $aggre
         $chunks = Factory::getChunkFactory()->filter([Factory::FILTER => $qF]);
         $aggregatedData["status"] = TaskUtils::getStatus($chunks, $keyspace, $keyspaceProgress);
       }
-
+      if (is_null($aggregateFieldsets) || in_array("totalNumberOfChunks", $aggregateFieldsets['task'])) {
+        $qF = new QueryFilter(Chunk::TASK_ID, $object->getId(), "=");
+        $aggregatedData["totalNumberOfChunks"] = Factory::getChunkFactory()->countFilter([Factory::FILTER => $qF]);
```
jessevz, line 219: You do a countfilter for the chunks, but potentially, chunks are already requested for the taskExtraDetails. I would put this in the same taskextardetails section after line 221. because this is all needed for the task detail page. Then you can just count the chunks that are already requested form the db. So not an extra database query is needed

PR #2118, opened 2026-05-20
Try to cover the tests with phpstan too

s3inlc:


Eric-Wasson (approved):
Checked it out on my machine and everything works (phpstan analyse only takes a few seconds)

PR #2150, opened 2026-05-22
Completed AbstractModelFactory tests

s3inlc:


ci/phpunit/dba/AbstractModelFactoryTest.php
```diff
@@ -340,4 +719,448 @@ public function testTimeseriesFilter(): void {
...
+  }
+  
+  /**
+   * Tests the columnFilter function which returns an array of values of the given column of matching rows
+   *
+   * @return void
+   * @throws Exception
+   */
+  public function testColumnFilter(): void {
+    $isSalted = random_int(1, 100);
```
Eric-Wasson: Shouldn't it be random_int(2, 100)? Otherwise there is a small chance that $isSalted is 1 and then this check fails: $this->assertSame([$hashlist_1->getId(), $hashlist_3->getId()], $ids);

Eric-Wasson (changes requested):

PR #2169, opened 2026-05-27
Correctly use task object to aggregate task information

s3inlc:
For the aggregations instead of the task ID, the taskwrapper ID was used, causing wrong values for the resulting data.


s3inlc:
the phpstan issues are fixed in PR #2170

jessevz (approved):

PR #2214, opened 2026-06-12
Show task owner (username) in tasks table

levensonblaine:
Adds an "Owner" column to the tasks list showing the username of the user who created each task/supertask.

The schema previously had no task–user relation, so this introduces one:

TaskWrapper gets a nullable userId column (install schema src/install/hashtopolis.sql, DBA model/factory, generator.php) with FK to User.
The creator's userId is set at all 6 new TaskWrapper(...) creation sites (web task creation, TaskUtils::createTask, pretask/supertask runs, applyPreconfTasks, rule-split inherits the original wrapper's userId). PretaskUtils::runPretask and SupertaskUtils::runSupertask gained an optional $user param, passed from web handlers, user-api, and apiv2 callers.
UserUtils::deleteUser nulls TaskWrapper.userId (same pattern as Agent.userId).
Util::loadTasks resolves usernames (single user query, map lookup) into taskOwner; templates tasks/index|normal|supertask render the new column (DataTable column indices shifted accordingly). Tasks without an owner (e.g. pre-existing or server-generated) show an empty cell.

### DIFF
--- a/src/dba/models/TaskWrapper.class.php
+++ b/src/dba/models/TaskWrapper.class.php
@@ -12,8 +12,9 @@ class TaskWrapper extends AbstractModel {
   private $taskWrapperName;
   private $isArchived;
   private $cracked;
+  private $userId;
   
-  function __construct($taskWrapperId, $priority, $maxAgents, $taskType, $hashlistId, $accessGroupId, $taskWrapperName, $isArchived, $cracked) {
+  function __construct($taskWrapperId, $priority, $maxAgents, $taskType, $hashlistId, $accessGroupId, $taskWrapperName, $isArchived, $cracked, $userId = null) {
     $this->taskWrapperId = $taskWrapperId;
     $this->priority = $priority;
     $this->maxAgents = $maxAgents;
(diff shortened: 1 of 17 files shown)


s3inlc:
Hi, thanks for creating your pull request. Please note that we are currently heavily working on the dev branch to finalize v1.0.0 which will feature the new frontend and a new unified API for both the frontend and also for scripting. From this point on, the current frontend and the user-api will be deprecated and will be removed after some time.

Until this is release, we will not do any database changes (unless really needed) to avoid to have to implement this for all the current parts as well (which will then be removed a bit after the release). 
If you are interested, please check out how your addition of an owner or creator user for tasks would have to be added there.

Also, if you look at it from a perspective a bit further away, this may not be the only place it would be good to have some "audit" information about who did what. Maybe it could also be an idea to have this at quite some other objects as well. Or if you think further, e.g. when you also want to know if someone updated an object, it might make sense to have some tables outside of the objects being responsible to log who did what and when (allowing that you could audit different objects and reconstruct what was done).

PR #2250, opened 2026-06-22
Fix filter ACL returning duplicate elements for agents

### LINKED ISSUE #2233: [BUG]: filterACL not working
### Version Information

rc1

### Hashcat

_No response_

### Description

For access groups permissions we use filterACL function. This uses inner joins which can give back to much rows from the database when the user is part of multiple accessgroups.

Another issue is that some acl filters rely on a nulable field, making them dissapear if the field is null. This should probably be fixed by a left join

jessevz:
closes #2233

### DIFF
--- a/src/inc/apiv2/model/AgentErrorAPI.php
+++ b/src/inc/apiv2/model/AgentErrorAPI.php
@@ -4,6 +4,7 @@ namespace Hashtopolis\inc\apiv2\model;
 
 use Hashtopolis\inc\utils\AccessUtils;
 use Hashtopolis\dba\models\AccessGroupAgent;
+use Hashtopolis\dba\models\Agent;
 use Hashtopolis\dba\ContainFilter;
 use Hashtopolis\dba\models\Hashlist;
 use Hashtopolis\dba\JoinFilter;
@@ -15,6 +16,7 @@ use Hashtopolis\dba\models\User;
 use Hashtopolis\inc\apiv2\common\AbstractModelAPI;
 use Hashtopolis\inc\apiv2\error\HttpError;
 use Hashtopolis\inc\Util;
+use Hashtopolis\dba\ExistsFilter;
 
 
 class AgentErrorAPI extends AbstractModelAPI {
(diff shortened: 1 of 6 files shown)


src/inc/apiv2/model/AgentErrorAPI.php
```diff
@@ -56,14 +58,15 @@ protected function getFilterACL(): array {
     
     return [
       Factory::JOIN => [
-        new JoinFilter(Factory::getAccessGroupAgentFactory(), AgentError::AGENT_ID, AccessGroupAgent::AGENT_ID),
+        // new JoinFilter(Factory::getAccessGroupAgentFactory(), AgentError::AGENT_ID, AccessGroupAgent::AGENT_ID),
```
s3inlc: do we have to keep this commented out, can't we remove it?

src/inc/apiv2/model/AgentErrorAPI.php
```diff
@@ -56,14 +58,15 @@ protected function getFilterACL(): array {
...
         new JoinFilter(Factory::getTaskFactory(), AgentError::TASK_ID, Task::TASK_ID),
         new JoinFilter(Factory::getTaskWrapperFactory(), Task::TASK_WRAPPER_ID, TaskWrapper::TASK_WRAPPER_ID, Factory::getTaskFactory()),
         new JoinFilter(Factory::getHashlistFactory(), TaskWrapper::HASHLIST_ID, Hashlist::HASHLIST_ID, Factory::getTaskWrapperFactory()),
       ],
       Factory::FILTER => [
-        new ContainFilter(AccessGroupAgent::ACCESS_GROUP_ID, $accessGroups, Factory::getAccessGroupAgentFactory()),
+        // new ContainFilter(AccessGroupAgent::ACCESS_GROUP_ID, $accessGroups, Factory::getAccessGroupAgentFactory()),
```
s3inlc: do we have to keep this commented out, can't we remove it?

s3inlc (changes requested):

s3inlc (approved):

PR #2286, opened 2026-06-26
Upgrading to mysql:9.7 and test also upgrades with different versions

### LINKED ISSUE #2238: [ENHANCEMENT]: upgrade to mysql 9.7
### Description

Mysql 9.7 is the new lts. We should test and verify if hashtopolis is compatible with 9.7 and then change the mysql version in the docker compose

s3inlc:
closes #2238


jessevz (changes requested):
Also change the .devcontainer/docker-compose.mysql.yml to mysql 9.7 to make the devcontainer in sync

jessevz (approved):

PR #2290, opened 2026-07-01
check for task completion by using task progress

### LINKED ISSUE #2270: [BUG]:  Tasks show wrong status
### Version Information

rc-2

### Hashcat

_No response_

### Description

If you have a running task, the symbol of the task status switches to a green arrow once 100% of the task is distributed. One expects to have an organge button blinking in the process of distributing and only a green error once the task is 100% searched.

jessevz:
closes #2270

### DIFF
--- a/ci/phpunit/inc/utils/TaskUtilsTest.php
+++ b/ci/phpunit/inc/utils/TaskUtilsTest.php
@@ -46,9 +46,9 @@ final class TaskUtilsTest extends TestBase {
    */
   public function testGetStatus(): void {
     $taskObjects = $this->createTaskHelper();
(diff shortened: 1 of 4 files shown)


s3inlc (changes requested):
I agree with copilot on its comment about removing the check for progress < 10000, currently indeed the function only is called by filtering before, but the function may be used at some other place later.

What probably would be the best way is to integrate the query to get the chunks itself into the function getStatus(), so that you only pass the Task object as argument and everything else is done inside the function. This avoids that mistake calls can be made to the function (with not the right chunks filtered out) and also reduces code duplication.

s3inlc (approved):

PR #2296, opened 2026-07-02
Remove encoding of HTML entities by the API

### LINKED ISSUE #2278: [BUG]: Changing Forbidden characters causes fatal error
### Version Information

rc-2

### Hashcat

_No response_

### Description

Steps to reproduce the issue:

Go to the config settings and edit the forbidden characters. 

Default value:

Image

Now add any character e.g. "9". After updating and refreshing this appears:

Image

You would have to edit again. If you dont notice, you cannot create tasks anymore, because "#" is added to the forbidden characters.

novasam23:
Fixes #2278

### DIFF
--- a/ci/apiv2/test_config.py
+++ b/ci/apiv2/test_config.py
@@ -1,6 +1,5 @@
 from hashtopolis import Config
 from utils import BaseTest
-from hashtopolis.hashtopolis import HashtopolisResponseError
 
 class ConfigTest(BaseTest):
     model_class = Config
--- a/ci/apiv2/test_config.py
+++ b/ci/apiv2/test_config.py
@@ -1,6 +1,6 @@
 from hashtopolis import Config
 from utils import BaseTest
-
+from hashtopolis.hashtopolis import HashtopolisResponseError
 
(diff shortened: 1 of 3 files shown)


s3inlc:
Maybe it would make sense to write a pytest which tests creating and/or patching some object with some umlauts or other characters which would be affected by this htmlentities to verify for the future that it does not get added/changed by accident.

novasam23:
I've added two more tests, each of which should by itself have caught the underlying issue with this bug.

There is still lot's of uses of `htmlentities()`, especially within the utils, that should be removed. I'd suggest to do this after the 1.0.0 release.

s3inlc:
> I've added two more tests, each of which should by itself have caught the underlying issue with this bug.
> 
> There is still lot's of uses of `htmlentities()`, especially within the utils, that should be removed. I'd suggest to do this after the 1.0.0 release.

AFAIK all other uses of htmlentities() are mostly tied to the old frontend or at least due to it, so it will be easy to remove them when we start removing it.

jessevz (approved):

PR #2321, opened 2026-07-08
Also make task status finished when hashlist has been cracked

### LINKED ISSUE #2304: [ENHANCEMENT]: cracked task not completed status
### Description

Currently when a task cracks a hash but is not 100% searched, it will not have the completed status with a checkmark (but it will be green). For an user it is more clear if these tasks also have the completed status, this should be done in the backend

jessevz:


s3inlc:
You need to make sure, that this change is consistently done, e.g. not other places may define a different status. For example, at least here it would also need to be changed so that it also is the correct status on Task details, etc.:https://github.com/hashtopolis/server/blob/master/src/inc/utils/TaskUtils.php#L134

jessevz:
> You need to make sure, that this change is consistently done, e.g. not other places may define a different status. For example, at least here it would also need to be changed so that it also is the correct status on Task details, etc.:https://github.com/hashtopolis/server/blob/master/src/inc/utils/TaskUtils.php#L134

Yes I noticed that, but currently this function is only used for sub tasks of a supertask. And I felt it may not be very clear if all subtask will be on completed when a single subtasks cracks the hash. Maybe it is more clear to the user if only the cracked subtask will go to completed.

s3inlc:
> And I felt it may not be very clear if all subtask will be on completed when a single subtasks cracks the hash.

But this is exactly what will happen with your aggregation as well.. the hashlist may be cracked by any other task than the supertask you aggregate on and it will mark the supertask with a green check (even if there was not crack at all in that supertasks subtask).

jessevz:
> > And I felt it may not be very clear if all subtask will be on completed when a single subtasks cracks the hash.
> 
> But this is exactly what will happen with your aggregation as well.. the hashlist may be cracked by any other task than the supertask you aggregate on and it will mark the supertask with a green check (even if there was not crack at all in that supertasks subtask).

Yes I understand, but I feel like that is more clear, since you already see all the subtasks together, making the cracked sub task pop out more with the green checkmark.
But I don't mind adding the status to the subtasks aswell. I just hope it will still be understandable for the end user

jessevz:
Preferably we might even want a 4th state were say, task has already been cracked by a different task, but I am not sure how to even signify that in the UI with a good icon

s3inlc (approved):

s3inlc:
will you develop the changes for a new state ('skipped' or similar) in this branch, or will we discard this one?

jessevz:
@s3inlc made changes by adding a skipped state, it is now ready for review again

s3inlc (approved):

PR #2329, opened 2026-07-14
Add aggregate field crackingTime to AgentAssignment

### LINKED ISSUE #2297: [ENHANCEMENT]: add aggregation on AgentAssignment
### Description

Make it possible that an aggregation is possible on the AgentAssignment model called `timeSpent` providing the time an agent has worked on that task (i.e. this means summing up the differences between the two timestamps on all chunks of this task which the agent worked on). This can be implemented pretty similar to the `timeSpent` aggregation on agent just with the difference that it's not on all chunks, but just of the specific task.

novasam23:
Fixes #2297


novasam23:
A few things to add:

- For the `AgentAssignment` I suggest to call the aggregate `crackingTime` instead of `timeSpent` in order to be consistent with how it's called on the Agent model

- A similar calculation is also done in `TaskUtils` called `getTimeSpentOnTask`. However, since this involves potentially multiple agents which results in parallel time spans, the calculation is more complex and hence done entirely in the API layer. A future enhancement might be, to move this to the DB for performance reasons, but I think that would require making the DB access classes capable of DB window functions. Definitely something for the future, if at all.

- As for the existing aggregate on the `Agent`, currently `crackingTime` is implemented in the `AgentAPI`. One benefit of having that in a utilities class (similar to `getTimeSpentOnTask`) would be to better test the "business logic" of correctly calculating the time values via PHP tests. Another benefit would be reducing the occurrences of duplicate code, since we now have almost identical calculations for the `Agent` and `AgentAssignment`.

- In short, I'm proposing to move this calculation to `TaskUtils` and consolidate existing occurrences accordingly.

- With window functions in the DB access layer, this could be further consolidated into a single generic time span (i.e. solveTime minus dispatchTime) aggregating function, to be used even by the Task entity.

Tests

- Talking about tests, I did implement a Pytest, but chunk creation and manipulation is very restricted via API, hence it's difficult to create meaningful test cases that cover all sorts of chunk processing constellations.

- The more relevant test I think is the PHP test, that creates different chunks with different timing constellations, assigned to multiple agents, etc.

s3inlc:
I agree with your consolidation, but I think it would be better to move it to AgentUtils and not TaskUtils, as it's agent dependent (and Task only in certain cases). Especially, as this way of do the diff of the sums only works when calculating for one agent, but not if for example we calculate the time which has been spent on task (wall clock time).

novasam23:
@s3inlc I moved the code for the cracking time aggregation to the `AgentUtils` as you suggested, I agree that this class is a better fit.

I also refactored the PHP unit test code accordingly to correspond with this change.

s3inlc (approved):

PR #2349, opened 2026-07-17
Add more aggregate fields to the agent assignment model

### LINKED ISSUE #2339: [ENHANCEMENT]: Further aggregate fields required for AgentAssignment
Related to #2329 and #2280 

Assigned agents view in task details page requires more aggregate fields: speed, cracked, keyspace searched

novasam23:
Closes #2339

### DIFF
--- a/src/inc/apiv2/model/AgentAssignmentAPI.php
+++ b/src/inc/apiv2/model/AgentAssignmentAPI.php
@@ -115,7 +115,6 @@ class AgentAssignmentAPI extends AbstractModelAPI {
         'cracked' => [$this, 'getAggregateCracked'],
         'currentSpeed' => [$this, 'getAggregateCurrentSpeed'],
         'searched' => [$this, 'getAggregateSearched'],
-        'cprogress' => [$this, 'getAggregateCProgress'],
       ]
     ];
   }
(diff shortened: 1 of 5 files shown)


s3inlc (changes requested):
In the old frontend, it was displaying the number of the keyspace covered plus the percentage. Ideally, we can provide this for the new frontend as well, but in order to allow this, the 'searched' aggregation should not return the percentage, but just the value of the keyspace, so the frontend can then calculate how many percent this is of the whole task (i.e. just removing the wrapping `Util::showperc()`).

novasam23:
I've added the additional aggregate field as you suggested

s3inlc:
Sorry, maybe I was not clear. I meant to update the 'searched' aggregate. I don't think it makes sense to have both cprogress and searched as they're quite redundant, except just the percentage calculation. But I think in general we should avoid percentage calculations in the backend, as this is a frontend part (displaying) and it's more accurate to deliver the actual number and leave it to the frontend (or a potential API user) to decide what they do with the value.

novasam23:
Totally agree, I think the API should not be concerned with the formatting of data, that's the UI's concern. But to keep it consistent, I would suggest to get rid of all those aggregate fields `searched` that currently exist in some API model endpoints. We should just keep the `cprogress` field, which is the raw value.

novasam23:
Removing the existing occurrences of `searched` would also require modifications in the UI though. Still better to get this change in before the 1.0.0 release, because removing fields **after** they existed in a major release at one point is not ideal.

s3inlc (approved):

PR #2373, opened 2026-07-27
Explanation of custom themes for docs

### LINKED ISSUE #1688: [ENHANCEMENT]: Write documentation for current permissions
### Description

We currently have a weird behavior in the permissions in the new UI. This is because we still have to handle the permissions in the old ui. THis should be explained and mentioned in the docs, maybe in the FAQ?

ObsidianOracle:
I added the description for using the custom css themes to the docs in the faq section.


s3inlc:
For a FAQ question this answer is way too long, also it duplicates quite a bit written already in the explanation in the frontend. I would recommend to give a very short answer (basically "yes it's possible" and maybe some few sentences) and then refer to https://github.com/hashtopolis/web-ui/tree/master/custom-themes.
If there is more detailed explanation needed than currently available there, add these additional steps directly into the files in the frontend.

Alternatively, if we like to have all this documented in the docs, then we should add it to the normal docs (everything from the theme folder in the frontend and remove it there).

s3inlc:
We could place it under advanced install as a new subsection there: https://docs.hashtopolis.org/installation_guidelines/advanced_install/

ObsidianOracle:
changed to a small answer and refered to the repo.

correct-horse-battery-bench:
@ObsidianOracle We might also want to link from the repo to the docs and move the content there? If the docs are source of truth and if a developer wants to know more they will see the reference from the repo readme to the docs.

ObsidianOracle:
Thanks @correct-horse-battery-bench. Its already in the repo readme. Have a look if we should adjust here and there.

doc/faq_tips/faq.md
```diff
@@ -265,6 +266,11 @@ No.
 
 It is not officially supported and there are no pre-built docker images available. ARM builds must be custom made.
 
+---
+<span style="font-size:1.2em; font-weight:bold;">❓ Can I use a custom layout and how can I embed it?</span>
```
s3inlc: In the doc we have the terms `custom theme`, `customize web interface` and here the term `custom layout` is used (the only place overall the theme docs). Why not just write `custom theme` here as well, to keep some useful nomenclature. 
Also `layout` probably can be misunderstanding anyway, as this could mean that objects could be arranged differently etc.

doc/faq_tips/faq.md
```diff
@@ -265,6 +266,11 @@ No.
...
+
+Yes. You find all the necessary steps documented in our advanced installation guidelines.
```
s3inlc: I would directly refer to the relevant doc section with a markdown link.

doc/installation_guidelines/custom-css.md
```diff
@@ -0,0 +1,256 @@
+# Customize Web Interface
```
s3inlc: I would do the overall title `Custom Themes` (as that's basically what is covered here) and again not use a different term for the same.

And then the next section should then probably be named `Using Custom Themes` (there it is explained how you could use existing ones) and the other level 2 section then is `Custom Theme Authoring Guide` (there it is explained how new themes can be created).

doc/installation_guidelines/custom-css.md
```diff
@@ -0,0 +1,256 @@
...
+
+If you would like to customize the web-interface using your own layout just follow the instructions explained here:
```
s3inlc: I would phrase `customize the web-interface` and `own layout` differently. As stated in other comments the words could be misleading.

Proposal: `In the frontend it is possible to plug in themes. This guide covers how you can install such themes and how to create a custom one.`

doc/installation_guidelines/custom-css.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Custom Themes
```
s3inlc: adjust title, see comment on line 1

doc/installation_guidelines/custom-themes.md
s3inlc, line 1: call the file `custom-themes.md`, css is too technical and does not really tell what is inside this doc.

doc/installation_guidelines/custom-css.md
```diff
@@ -0,0 +1,256 @@
...
+
+If you would like to customize the web-interface using your own layout just follow the instructions explained here:
+
+## Custom Themes 
+
+Drop `.css` files in this folder to register custom runtime themes.
+
+No additional JSON or TypeScript changes are required for standard theme creation.
+
+For a complete authoring guide, see **Custom Theme Authoring Guide**.
```
s3inlc: either link down to the section or maybe just add the word `below` at the end.

doc/installation_guidelines/custom-css.md
```diff
@@ -0,0 +1,256 @@
...
+- A theme whose CSS declares `color-scheme: dark` is flagged dark, so the app uses its dark-mode logo, icon colors, and chart palettes.
+
+### Reserved IDs
+
+These ids are built-in and cannot be overridden by custom files:
+
+- `light`
+- `dark`
+
+### How to get help generating a theme?
```
s3inlc: I think this subsection can be omitted as now the creation guide is just right below.

s3inlc (changes requested):

doc/faq_tips/faq.md
```diff
@@ -265,6 +266,25 @@ No.
...
 
+---
+<span style="font-size:1.2em; font-weight:bold;">❓ Can I use a custom theme and how can I embed it?</span>
+
+Yes. You find all the necessary steps documented [here](https://docs.hashtopolis.org/installation_guidelines/custom-theme/).
+
+---
+<span style="font-size:1.2em; font-weight:bold;">❓ Configuring my global permissions leads to a weird behaviour: Some permissions are disabled/enabled once I configure another permission. Is this normal?</span>
+
+Yes, this behavior is completely normal and actually intentional. With the 1.0 release of Hashtopolis, the permission scheme from the old UI was migrated into the new UI, allowing dependencies between individual permissions to be defined and adjusted much more granularly.
```
s3inlc: I would add and highlight here, that this is done to allow the sync between the old and new frontend which are currently in parallel, and that this behavior will be removed once the old frontend will get removed.

doc/faq_tips/faq.md
```diff
@@ -265,6 +266,25 @@ No.
...
+
+**Why permissions change automatically**
+When you configure a specific permission, the system may automatically enable or disable other, dependent permissions. This is a core part of the new granular permission model, not a bug.
```
s3inlc: These changes are not a core part of the new granular model. It's only due to these above mentioned legacy reason. I would drop this section here.

doc/faq_tips/faq.md
```diff
@@ -265,6 +266,25 @@ No.
...
+
+**Recommendations for everyday use**
+Avoid manually overriding automatic enabling or disabling of permissions, as this can lead to unexpected behavior in your setup.
```
s3inlc: Not sure what this really means to a user? As at some point you have to change permissions manually, but then maybe just the user needs to know that he has to accept certain combinations because it's not possible to create certain combinations.

s3inlc (changes requested):

s3inlc (approved):

PR #2378, opened 2026-07-28
Agent API testing

s3inlc:
Tests with pytest to cover the existing API used by the agents to be able to detect at least obvious issues automatically.

Two bugs had to be fixed while writing the tests (and branch is based off the protocol fix branch, as these were issues as well).

Also the teardown test running (to check if all test data was deleted) had another bug where it checked with > 1 instead of > 0 which caused it still to pass when one object was not correctly deleted.


ci/apiv2/test_agent_protocol.py
```diff
@@ -0,0 +1,1774 @@
...
+class TestSendProgress(AgentProtocolBase):
+    """Tests for the sendProgress action — agent reports chunk progress, cracked hashes, and receives zaps."""
+
+    CRACK_HASH = "cc03e747a6afbbcbf8be7668acfebee5"
+    CRACK_PLAIN = "test123"
+    CRACK_HEX = "74657374313233"
+    FAKE_HASH = "ffffffffffffffffffffffffffffffff"
+    CRACK_HASH_2 = "098f6bcd4621d373cade4e832627b4f6"
+    TWO_HASH_SOURCE_DATA = (
+        "Y2MwM2U3NDdhNmFmYmJjYmY4YmU3NjY4YWNmZWJlZTUKMDk4ZjZiY2Q0NjIxZDM3M2NhZGU0ZTgzMjYy"
```
jessevz: this looks a bit weirdly. I understand that python will concatenate line 1279 and 1280 and then two hashes are base64 encoded in a single base64 encoded string. But i think it will look more cleanly and clear if you split the total base64 string in 2 equal parts. now it kinda looks like "N2I0ZjYK" is the second hash base64 encoded

jessevz (approved):

PR #2383, opened 2026-07-29
Add permission behaviour to faq

ObsidianOracle:
I added the dependent behavior of the permissions to the faq sectionn


ObsidianOracle:
Included in https://github.com/hashtopolis/server/pull/2373

Issue #2295, opened 2026-07-01
[ENHANCEMENT]: adaptively alter speed benchmark


jessevz:
### Description

We notice that certain tasks will consistently give a wrong speed benchmark, making the chunks way to short when speed benchmark is used. It would be great if the speed benchmark can be adapted when the finished chunk time has deviated a lot from the configured chunk time in the task. a solution could be when a chunk has completed and a big deviation is detected, the speed benchmark value in the assignment is adjusted. Another solution could be to alter the chunk time in the task, but this feels a bit hacky

s3inlc:
Can you more specifically give examples where it is off (type of tasks etc.)?
There is an ongoing problem in hashcat, that certain benchmark results (especially with salted hashes) are quite off: https://github.com/hashcat/hashcat/pull/4637

Also, there is a quite old PR open which tried to address such cases: #729 
The reason this one was never merged is that it would also shrink chunks which is not desired due to risk of problems.

PR #2427, opened 2026-08-07
Implement adaptive chunk sizes / adaptive speed benchmark results

### LINKED ISSUE #551: Adaptive chunk sizes

mr-creditcard:
This PR is a continuation of the work done by @0xVavaldi in #729. I have cherry-picked their commits and added my own on top in an attempt to keep attribution in place. It addresses issues #551 & #2295.

The aforementioned PR has a number of requirements set by @s3inlc in the comments which have remained unaddressed thusfar, this PR aims to implement these. The changes include:

- Shrinking chunks: Because of explanations given in comments https://github.com/hashtopolis/server/pull/729#issuecomment-999093298 & https://github.com/hashtopolis/server/pull/729#issuecomment-1049176267 we do not want to shrink chunk sizes. This PR addresses this in commit eee602a92ed3a2d353bd8a8ba864959f400198ac by early-returning if the multiplier used to adjust the benchmark score is < 1.
- Code organization: The original PR added code directly into `APISendProgress.php`'s switch statement over `DHashcatStatus`. In commit dbd4b326ebf01e7d8d2f9d10590c7f81e456fd7c this logic is moved to `TaskUtils.php` instead and is called with a function call.
- Global option to disable functionality: The original PR did not have a global config option to disable the behaviour introduced by this PR. A config option is added in commit https://github.com/hashtopolis/server/commit/bc671d2835c65ef6a625fd54fc0c8add46663459.

Did some quick tests with this code and it seems to work fine. Would appreciate someone else testing this as well though!

Closes https://github.com/hashtopolis/server/issues/551 & https://github.com/hashtopolis/server/issues/2295.

### DIFF
--- a/src/inc/utils/TaskUtils.php
+++ b/src/inc/utils/TaskUtils.php
@@ -44,6 +44,7 @@ use Hashtopolis\inc\defines\DPayloadKeys;
 use Hashtopolis\inc\defines\DPrince;
 use Hashtopolis\inc\defines\DTaskStaticChunking;
 use Hashtopolis\inc\defines\DTaskTypes;
+use Hashtopolis\inc\defines\DServerLog;
 use Hashtopolis\inc\handlers\NotificationHandler;
 use Hashtopolis\inc\HTException;
 use Hashtopolis\inc\SConfig;
@@ -1518,5 +1519,37 @@ class TaskUtils {
     $qF4 = new QueryFilter(Chunk::PROGRESS, 10000, "<");
     $agg = new Aggregation(Chunk::SPEED, Aggregation::SUM);
     return (int)(Factory::getChunkFactory()->multicolAggregationFilter([Factory::FILTER => array_filter([$qF1, $qF2, $qF3, $qF4])], [$agg])[$agg->getName()] ?? 0);
+  }
+
+  /**
+   *  Adjust the benchmark of an agent for a task to ensure the actual time it takes to calculate a chunk
+   *  is within 20% of the configured chunk duration time.
+   *
+   * @param Chunk $chunk
+   * @param Task $task
+   * @throws Exception
+   */
+  public static function tuneChunkDuration(Chunk $chunk, Task $task, Agent $agent): void {
+    $timeTaken = $chunk->getSolveTime() - $chunk->getDispatchTime();
+    if($timeTaken < 0) return; // prevent math & logic errors
+
+    $differenceToChunk = $task->getChunkTime() / $timeTaken;
+    // Limit how much difference a chunk can have to the previous.
+    $differenceToChunk = ($differenceToChunk > 1.5) ? 1.5 : $differenceToChunk;
+    $differenceToChunk = ($differenceToChunk < (2/3)) ? (2/3) : $differenceToChunk;
+    if ($differenceToChunk > 0.8 && $differenceToChunk < 1.2) return;
+    
+    $qF1 = new QueryFilter(Assignment::AGENT_ID, $chunk->getAgentId(), "=");
+    $qF2 = new QueryFilter(Assignment::TASK_ID, $chunk->getTaskId(), "=");
+    $assignment = Factory::getAssignmentFactory()->filter([Factory::FILTER => [$qF1, $qF2]])[0];
+
+    $benchmark = $assignment->getBenchmark();
+    $benchmarkParts = explode(":", $benchmark);
+    if($benchmarkParts[0] == 0 || count($benchmarkParts) != 2) return;
+    $newBenchmark = $differenceToChunk * $benchmarkParts[0];
+    $assignment->setBenchmark(round($newBenchmark).":".round($benchmarkParts[1]));
+    DServerLog::log(DServerLog::INFO, "{$timeTaken}---{$task->getChunkTime()}", [$agent, $assignment]);
+    DServerLog::log(DServerLog::INFO, "Multiplied the benchmark of agent by ".round($differenceToChunk,2), [$agent, $assignment]);
+    Factory::getAssignmentFactory()->update($assignment);
   }
 }
--- a/src/inc/utils/TaskUtils.php
+++ b/src/inc/utils/TaskUtils.php
@@ -1537,7 +1537,9 @@ class TaskUtils {
     // Limit how much difference a chunk can have to the previous.
     $differenceToChunk = ($differenceToChunk > 1.5) ? 1.5 : $differenceToChunk;
     $differenceToChunk = ($differenceToChunk < (2/3)) ? (2/3) : $differenceToChunk;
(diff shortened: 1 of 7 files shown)


src/migrations/postgres.1/20260810144313_chunk-auto-tune-config.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO `Config` (`configId`, `configSectionId`, `item`, `value`) VALUES
```
jessevz: I think some of the tests are failing because of the backticks, this is mysql syntax and not postgres syntax

src/inc/utils/TaskUtils.php
```diff
@@ -1519,4 +1520,42 @@ public static function getCurrentSpeedOfTask(int $taskId, ?int $agentId = null):
...
+    if ($differenceToChunk < 1.2) return; // if the difference is less than 20% don't make any adjustments, margin of error. We also disregard any 
+                                          // adjustments that would result in a smaller chunk size (anything < 1.0), since we do not want to perform
+                                          // automatic reductions in chunk size.
+    
+    // Limit the multiplier to 1.5, this prevents overshooting the time an agent should work on a chunk.
+    // This value could be increased (to allow for faster ramping up), but since the benchmark results and/or 
+    // chunk calculation times for small chunks can be pretty inaccurate this is a pretty safe and reasonable
+    // multiplier. Keep in mind the chunk duration will not be tuned down, only up. So once overshot, you're
+    // stuck with that time for the remainder of the task.
+    $differenceToChunk = ($differenceToChunk > 1.5) ? 1.5 : $differenceToChunk;
```
jessevz: bit nitpicky of me but the magic number 1.5 would be cleaner in a variable ;)

jessevz (changes requested):
Code looks really good, apart from the small syntax error phpstan picked up and the postgres syntax it looks great. I have not tested the code yet but once these things are fixed I can do a test run. Maybe @s3inlc also want to review this

src/migrations/postgres.1/20260810144313_chunk-auto-tune-config.sql
mr-creditcard: Right, forgot about that one! I'll get on fixing that, looking back at it I may have also used the wrong directory for the migration, but I will double-check that as well.

src/inc/utils/TaskUtils.php
mr-creditcard: That is indeed cleaner, I've moved the two magic numbers to their own variables now!

src/migrations/postgres.1/20260810144313_chunk-auto-tune-config.sql
mr-creditcard: Updated the migration to what I believe is the correct syntax and moved it to the correct directory (the not-`.1` directory).

mr-creditcard:
Rebased this PR on master!

src/inc/utils/TaskUtils.php
```diff
@@ -1533,4 +1534,47 @@ public static function getCurrentSpeedOfTask(int $taskId, ?int $agentId = null):
...
+
+  /**
+   *  Adjust the benchmark of an agent for a task to ensure the actual time it takes to calculate a chunk
+   *  is within 20% of the configured chunk duration time.
+   *
+   * @param Chunk $chunk
+   * @param Task $task
+   * @throws Exception
+   */
+  public static function tuneChunkDuration(Chunk $chunk, Task $task, Agent $agent): void {
```
s3inlc, line 1546: in order to keep consistent with the code styling, we would like to avoid having in-line if statements, could you adjust the three occurrences in this file with a {} wrapped statement?

s3inlc (changes requested):
Thanks for creating the PR, there is just some minor thing to adjust, otherwise this should be fine. Though we will have to adjust also in the frontend, so that the config variable can be changed there, so there may be some delay in merging it in here until the frontend also is ready with the change.

four-random-common-words:
@s3inlc Frontend PR for this: https://github.com/hashtopolis/web-ui/pull/823

src/inc/utils/TaskUtils.php
mr-creditcard: Done!

src/inc/utils/TaskUtils.php
```diff
@@ -1533,4 +1534,52 @@ public static function getCurrentSpeedOfTask(int $taskId, ?int $agentId = null):
...
+    $timeTaken = $chunk->getSolveTime() - $chunk->getDispatchTime();
+    if($timeTaken < 0) {
```
s3inlc: I think this should be <= 0, otherwise the division below would cause issues in case $timeTaken == 0

s3inlc (changes requested):
Thanks for your adjustments, sorry I missed the <0 comparison in the first review, beside this I think it looks good to merge.
As part of the merge it is needed to adjust the migrations filenames to be adjusted to the current date and time (to keep script order for migrations in master). Can you just set it to an actual date and time together with the small change and then I'll merge shortly after.

src/inc/utils/TaskUtils.php
mr-creditcard: Fixed this, thanks!

mr-creditcard:
> [...] As part of the merge it is needed to adjust the migrations filenames to be adjusted to the current date and time (to keep script order for migrations in master). Can you just set it to an actual date and time together with the small change and then I'll merge shortly after.

Did that just now!

s3inlc:
> Did that just now!

Perfect, thank you! Sorry that I didn't trigger the workflows earlier, there seems to be a syntax issue with the postgres migration (I think the single quotes are not valid, probably no quotes should be used on the column names). And also PHPstan highlighted something (should be `$agent` instead of `$this->agent` and probably wrapping intval() around the benchmark value should do the trick.

PR #2430, opened 2026-08-10
2391 bug filtering for active chunk on agent aggregate is not ideal

### LINKED ISSUE #2391: [BUG]: Filtering for active chunk on agent aggregate is not ideal
### Version Information

dev

### Hashcat

_No response_

### Description

In the AgentAPI for filtering the active chunk only the agentId and status (=RUNNING) are tested, which may not be completely ideal. There should also be a filter for completeness to avoid having chunks with some still running state but technically being completed (i.e. progress < 10000 and order by solveTime DESC).

novasam23:
Fixes #2391


src/inc/utils/AgentUtils.php
```diff
@@ -644,4 +645,24 @@ public static function getAggregateCracked(int $agentId, ?int $taskId = null): i
...
+   *
+   * @param int $agentId
+   * @param int|null $taskId
+   * @return Chunk|null
+   * @throws Exception
+   */
+  public static function getActiveChunk(int $agentId, ?int $taskId = null): ?Chunk {
+    $qFs = [];
+    $qFs[] = new QueryFilter(Chunk::AGENT_ID, $agentId, "=");
+    $qFs[] = $taskId !== null ? new QueryFilter(Chunk::TASK_ID, $taskId, "=") : null;
```
jessevz: This line looks a bit like a code smell to me, this will add null to the qFs array when when taskId is null. But we dont really need that null value. I think it is better to use a normal if statement instead of a tenary to only add this to the qFs[] when taskId is not null and otherwise dont add anything

jessevz: Ah I see now that you filter the null value away later by doing array_filter(). Still for readability I think it is better to just do a normal if statement

jessevz (changes requested):

jessevz (approved):

PR #2443, opened 2026-08-18
2404 stack 1 openapi generator and lint

correct-horse-battery-bench:
Refactor openapi spec generator to be in its own module, output openapi 3.1.0 version output and make sure it can generate the spec without needing a live db or application just from the registered model classes.


src/inc/apiv2/common/ApiRegistry.php
```diff
@@ -0,0 +1,150 @@
...
+    SpeedAPI::class,
+    SupertaskAPI::class,
+    TaskAPI::class,
+    TaskWrapperAPI::class,
+    TaskWrapperDisplayAPI::class,
+    UserAPI::class,
+    VoucherAPI::class,
+  ];
+
+  /** @var list<class-string> */
```
s3inlc: Similar as the typing for the model classes, the typing could be tighter for the helpers as well and set to:
```
/** @var list<class-string<AbstractHelperAPI>> */
```

s3inlc (changes requested):

s3inlc (approved):

PR #2447, opened 2026-08-18
2404 stack 4 model definition accuracy

correct-horse-battery-bench:
Update model definitions so the openapi spec becomes more accurate.

- make sure post fields are required as well as the same model is used for get and post
  -  has the effect that create needs to pass some fields which have defaults but this makes the openapi spec more accurate for gets


src/dba/models/NotificationSetting.php
```diff
@@ -39,9 +39,9 @@ function getKeyValueDict(): array {
...
+    $dict['objectId'] = ['read_only' => True, "type" => "int", "subtype" => "unset", "choices" => "unset", "null" => True, "pk" => False, "protected" => True, "private" => False, "alias" => "objectId", "public" => False, "dba_mapping" => False];
+    $dict['notification'] = ['read_only' => False, "type" => "str(50)", "subtype" => "unset", "choices" => ["ChatBot" => "ChatBot", "Discord Webhook" => "Discord Webhook", "Email" => "Email", "Example" => "Example", "Slack" => "Slack", "Telegram" => "Telegram", ], "null" => False, "pk" => False, "protected" => False, "private" => False, "alias" => "notification", "public" => False, "dba_mapping" => False];
```
s3inlc: Baking the existing Notifications that strongly into the model is probably not ideal. Users can put in their own notification classes and use them. With this change, this would require to change DBA model files

src/inc/utils/AgentUtils.php
```diff
@@ -390,12 +390,13 @@ public static function deleteDependencies(Agent $agent): bool {
...
    * @param int $taskId
    * @param User $user
+   * @param string $benchmark the benchmark to record for the assignment, "0" when none is known yet
    * @return ?Assignment
    * @throws HTException
    * @throws HttpError
    * @throws Exception
    */
-  public static function assign(int $agentId, int $taskId, User $user): ?Assignment {
+  public static function assign(int $agentId, int $taskId, User $user, string $benchmark = "0"): ?Assignment {
```
s3inlc: I think it would be better to defaul to empty string than 0.

s3inlc (changes requested):

s3inlc (approved):

PR #2480, opened 2026-09-08
Added the hashlist isSecret field to the taskdisplayview

### LINKED ISSUE #2462: [BUG]: hashlist secret unclear
### Version Information

1.0.0

### Hashcat

_No response_

### Description

In the new UI, hashlist are on default created as secret, this differs from the old UI. This part is fixed by: https://github.com/hashtopolis/web-ui/pull/811

In addition it would be nice if on the task overview page a lock is shown on secret hashlists, this is missing form the new UI and can make it more clear to the user why a task is not picked up

jessevz:
Added the issecret field from the hashlist to the taskdisplayview so that the frontend can render that it is a secret hashlist


s3inlc (approved):
Looks good, remember that we have to update the migrations datetime before merge.

PR #2498, opened 2026-09-17
Handling both benchmark default values properly

s3inlc:
The server was only handling the default value of '0' properly, but should also be able to treat '' (empty string) correctly.


four-random-common-words (approved):
LGTM 👍

